PR #8, opened 2018-09-12
Added documentation on how to configure maven and gradle

DreierF:


karottenreibe:
please merge master, that should fix the build problem :-)

agent/USAGE.md
```diff
@@ -171,6 +171,62 @@ This ensures that the performance of your application does not degrade.
 
 Please ask CQSE for special tooling that is available to instrument Java Web Start processes.
 
+## Store Commit to Manifest
+
+As it is very convenient to use the MANIFEST entries via `teamscale-commit-manifest-jar` to link artifacts to commits, 
+especially when tests are executed independently from the build.
```
karottenreibe: please mention that the following examples assume you're using Git

agent/USAGE.md
```diff
@@ -171,6 +171,62 @@ This ensures that the performance of your application does not degrade.
...
+### Gradle
+
+```groovy
+plugins {
+	id 'org.ajoberstar.grgit' version '2.3.0'
+}
+
+jar {
+	manifest {
+		attributes 'Branch': System.getProperty("branch")
```
karottenreibe: why not read that from git as well?

agent/USAGE.md
```diff
@@ -171,6 +171,62 @@ This ensures that the performance of your application does not degrade.
...
+		attributes 'Timestamp': grgit.log {
+			maxCommits = 1
+		}[0].dateTime.toInstant().getEpochSecond() * 1000L
```
karottenreibe: use .first() instead of [0] for clarity?
use toEpochMilli() instead of getEpochSecond * 1000?

agent/USAGE.md
```diff
@@ -171,6 +171,62 @@ This ensures that the performance of your application does not degrade.
...
+	}
+}
```
karottenreibe: did you test the script?

agent/src/main/java/eu/cqse/teamscale/jacoco/agent/AgentOptionsParser.java
```diff
@@ -222,9 +222,9 @@ private static CommitDescriptor getCommitFromManifest(File jarFile) throws Agent
 			Manifest mf = jarStream.getManifest();
```
karottenreibe: please rename to manifest to avoid cryptic abbreviations

agent/USAGE.md
DreierF: I wanted to show both ways, to be able to use it without having to google for it first and also because this will be executed mostly during the build, where there is no branch name available. AFAIK all build envs like Travis, GitLab, etc. checkout the specific commit in detached head mode, which means we can only rely on the BRANCH_NAME (or whatever it is called in the used tool) environment variable to determine the branch

agent/USAGE.md
DreierF: Good point

agent/USAGE.md
DreierF: yes

karottenreibe:
Please merge master to fix the conflicts

Issue #33, opened 2018-11-14
Document how to obtain commit timestamp for SVN


sewe:
Currently, we document how to obtain a commit timestamp suitable for Teamscale for Git (`git --no-pager log -n1 --format="%ct00`). We should do so for SVN as well.

I will prepare a pull request.

agoeb:
Fixed

PR #34, opened 2018-11-14
Document how to obtain commit timestamp for SVN

sewe:
This fixes issue #33.

- [x] [No new Teamscale findings](https://build.cqse.eu/teamscale/delta.html#input/jacoco-client/?showMergeFindings=true)
- [x] [Changes are tested adequately](https://build.cqse.eu/teamscale/tests.html#/jacoco-client/)
- [ ] USAGE.md updated in case of user-visible changes
- [ ] CHANGELOG.md updated
- [x] build.gradle version number increased according to semantic versioning

### DIFF
--- a/agent/README.md
+++ b/agent/README.md
@@ -70,10 +70,10 @@ echo `git rev-parse --abbrev-ref HEAD`:`git --no-pager log -n1 --format="%ct000"
   returns HEAD instead of the branch. In this case the environment variable provided by the build runner should be used 
   instead.
   
-  If **Subversion** is your VCS, you can get the timestamp part (but not the branch part) of the commit info via
+  If **Subversion** is your VCS and your reposiory follows the SVN convention with `trunk`, `branches`, and `tags` directories, you can get the commit info via
   
   ```bash
-date -d `svn info --show-item last-changed-date` "+%s000"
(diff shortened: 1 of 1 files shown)


agent/README.md
```diff
@@ -59,17 +59,23 @@ The following options are available:
...
+  
   Note: Getting the branch does most likely not work when called in the build pipeline, because Jenkins, GitLab,
   Travis etc. checkout a specific commit by its SHA1, which leaves the repository in a detached head mode and thus 
   returns HEAD instead of the branch. In this case the environment variable provided by the build runner should be used 
   instead.
   
+  If **Subversion** is your VCS, you can get the timestamp part (but not the branch part) of the commit info via
+  
+  ```bash
+date -d `svn info --show-item last-changed-date` "+%s000"
```
DreierF: I think it would be good to also have the command on how to get the branch name from the subversion repository here.

agoeb: Depending on the version of GNU date, you may also use +"%s%3N" for exact milliseconds. Otherwise, we might risk uploading the coverage *before* the code commit.

agoeb: Also, you might want to add the `-r revision` parameter.

sewe: > I think it would be good to also have the command on how to get the branch name from the subversion repository here.

That's tricky in SVN, as the branch name is just a convention. But it should be possible [to extract from the URL](https://stackoverflow.com/questions/5275924/what-svn-command-can-i-use-to-see-the-branch-im-currently-using). Will conduct some testing.

Issue #44, opened 2019-02-11
Support for Java 11


DreierF:
Requires update of JaCoCo to 0.8.3

DreierF:
Is available with version 11.2.0

PR #45, opened 2019-02-11
Updated JaCoCo to 0.8.3

DreierF:
- [x] [No new Teamscale findings](https://build.cqse.eu/teamscale/delta.html#input/jacoco-client/?showMergeFindings=true)
- [x] USAGE.md updated in case of user-visible changes
- [ ] CHANGELOG.md updated
- [x] build.gradle version number increased according to semantic versioning


DreierF:
Fixes #44

report-generator/src/main/java/org/jacoco/core/internal/analysis/CachingInstructionsBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+	public void fillCache() {
+		// Wire jumps:
+		for (final Jump j : jumps) {
+			j.wire();
+		}
+
+		// ADDED
+		// Traces back all instructions that are executed before reaching a probe
+		// and stores the mapping from probe to lines in #classCoverageLookup
+		for (final CoveredProbe p : coveredProbes) {
```
karottenreibe: If this is our code, please don't use unnecessary final and please don't use one-letter variable names

report-generator/src/main/java/org/jacoco/core/internal/analysis/CachingInstructionsBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+	 * ADDED
+	 * Helper to get the private field predecessor from an instruction.
+	 */
+	private Instruction getPredecessor(Instruction instruction) {
+		try {
+			Field f = instruction.getClass().getDeclaredField("predecessor");
+			f.setAccessible(true);
+			instruction = (Instruction) f.get(instruction);
+		} catch (NoSuchFieldException | IllegalAccessException e) {
+			e.printStackTrace();
```
karottenreibe: handle with proper logging? Also at least please log that this means we have a programming error that we have to handle here in the code (I think users can't do anything in this case, right?)

report-generator/src/main/java/org/jacoco/core/internal/analysis/CachingInstructionsBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+		}
+	}
+
+	/**
+	 * ADDED
+	 * Helper to get the private field predecessor from an instruction.
+	 */
+	private Instruction getPredecessor(Instruction instruction) {
+		try {
+			Field f = instruction.getClass().getDeclaredField("predecessor");
```
karottenreibe: please don't use one-letter variable names

report-generator/src/main/java/org/jacoco/core/internal/analysis/CachingInstructionsBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+				coveredLines.add(instruction.getLine());
+				instruction = getPredecessor(instruction);
+			}
+			classCoverageLookup.addProbe(p.probeId, coveredLines);
+		}
+	}
+
+	/**
+	 * ADDED
+	 * Helper to get the private field predecessor from an instruction.
```
karottenreibe: please explain what is stored in the predecessor field

report-generator/src/main/java/org/jacoco/core/internal/analysis/CachingInstructionsBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+	 * instances
+	 */
+	public void fillCache() {
+		// Wire jumps:
+		for (final Jump j : jumps) {
+			j.wire();
+		}
+
+		// ADDED
+		// Traces back all instructions that are executed before reaching a probe
```
karottenreibe: please explain why we need to trace back predecessors. is this because jacoco only inserts a probe at the end of a linear sequence of instructions?

report-generator/src/main/java/org/jacoco/core/internal/analysis/CachingInstructionsBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+	/**
+	 * Adds a new probe for the last instruction.
+	 *
+	 * @param probeId index in the probe array
+	 * @param branch  unique branch number for the last instruction
+	 */
+	void addProbe(final int probeId, final int branch) {
+//		REMOVED check of probes array and instead add the probes unconditionally
+//		final boolean executed = probes != null && probes[probeId];
+		currentInsn.addBranch(true, branch);
```
karottenreibe: is this line also added by us? if so please explain what this does and why it's needed. if not, please change the order of this and the next instruction and add a blank line to make it clear that this is unrelated to the comment

karottenreibe:
please update CHANGELOG
how did you test this? I think we should at least do some rudimentary manual test to make sure the caching still works as expected and the resulting coverage is not missing anything or has too much coverage
Or do we have automated tests for this?

report-generator/src/main/java/org/jacoco/core/internal/analysis/CachingInstructionsBuilder.java
DreierF: I converted into a RuntimeError as we cannot recover from this point anyway.

report-generator/src/main/java/org/jacoco/core/internal/analysis/CachingInstructionsBuilder.java
DreierF: Correct, added a short additional comment line.

DreierF:
We do have an automated smoke test in JaCoCoTestwiseReportGeneratorTest that still produces the same coverage as before.
However there is a known bug (TS-18106) that the coverage is not correct in all cases, but that will be fixed in a separate ticket. It only affects the line coverage numbers, but not TIA or TGA as a whole as we only consider method level coverage.

DreierF:
RFR
Is there anyway I can signal this via the GitHub UI?

karottenreibe:
Unfortunately, I don't know of any way. We could have a look at https://reviewable.io/ which maybe has a better review workflow than Github. But I haven't had the time to deep-dive into it to check it out. Just from the website I can't tell if this would work for us. If you'd like you can have a look and let me know what you think :-)

karottenreibe:
good work!

PR #52, opened 2019-03-06
(DON'T MERGE) TIA request forwarding

DreierF:
- [ ] CHANGELOG.md updated


agent/build.gradle
```diff
@@ -40,6 +40,7 @@ dependencies {
 
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.assertj:assertj-core:3.8.0'
+    testImplementation 'org.mockito:mockito-core:2.+'
```
DreierF: Please replace with a fixed version as variable versions make the build non-reproducible. Just ran into a major issue when updating mockito in TS today ;-)

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -33,8 +35,8 @@
 	protected final IXmlStore store;
 
 	/** Constructor. */
-	/*package*/ Agent(AgentOptions options) throws IllegalStateException, UploadStoreException {
-		super(options);
+	/*package*/ Agent(AgentOptions options, IAgent agent) throws IllegalStateException, UploadStoreException {
```
DreierF: Reame to `agentDelegate` or `jacocoAgent`? Looks strange if the agent gets an agent as parameter 😉

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
```diff
@@ -5,46 +5,117 @@
...
-		get("/test", (request, response) -> controller.getSessionId());
+	/**
+	 * Returns the lowest unused local port that is equal or greater than the
+	 * one provided in {@link #options}.
+	 */
+	private int determineFreePort() throws IOException {
+		for (int currentPort = options.getHttpServerPort(); currentPort < MAX_PORT_NUMBER;) {
+			try (ServerSocket s = new ServerSocket(currentPort)) {
+				// found a free port. Jump out
+				return currentPort;
```
DreierF: Do we already reserve the port by creating a ServerSocket here?
We will run into race conditions with other agents.
If the port is already reserved for us, then please document here.

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
```diff
@@ -71,21 +144,115 @@ private String handleTestStart(Request request, Response response) {
...
+
+	/** Stops the HTTP service and unregisters from the primary agent. */
 	@Override
 	protected void prepareShutdown() {
-		stop();
+		if (primaryAgent != null) {
+			try {
+				primaryAgent.unregister(getPort()).execute();
+			} catch (IOException e) {
+				logger.error("Unable to unregister from primary agent.");
```
DreierF: Should we pass the exception to the logger?

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
```diff
@@ -71,21 +144,115 @@ private String handleTestStart(Request request, Response response) {
...
+	 * Extracts and validates the port number from the request.
+	 * 
+	 * @throws IllegalArgumentException
+	 *             if the port number is missing, not a number, or an invalid
+	 *             port.
+	 */
+	private static int extractPortFromQuery(Request request) throws IllegalArgumentException {
+		String portString = request.queryParams("port");
+		if (portString == null || portString.isEmpty()) {
+			throw new IllegalArgumentException("Port is missing!");
```
DreierF: Should be BadRequestException, right?

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
```diff
@@ -71,21 +144,115 @@ private String handleTestStart(Request request, Response response) {
...
+		}
+		int port = Integer.parseInt(portString);
+		if (port < 1024 || port >= MAX_PORT_NUMBER) {
+			throw new NumberFormatException("Port " + port + " is not a valid port.");
```
DreierF: Same here

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
agoeb: This is a known issue. We do *not* reserve the port, because there is no way to do instantiate the spark service with an already established server socket. So we need to release it and hope that it is still available once the integrated Jetty tries to bind to the same port. Asked the customer to check whether this is an issue in practice. Will add a comment to the code to make that clear.

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
agoeb: Seems we don't have that one available. This is Teamscale specific. I could create one.

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
agoeb: I thought since `Integer.parseInt` already throws `NumberFormatException`, which extends `IllegalArgumentException` I use this here as well and can catch all of them at once in the calling method, since they are all related.

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
agoeb: Will do

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
agoeb: done

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
agoeb: Also added a hard fail, so we do not lose data.

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
```diff
@@ -234,9 +237,15 @@ private String error(Response response, String message, Object... arguments) {
...
-	private void logInitException(Exception e) {
+	/**
+	 * Logs errors that occurred during agent initialization. Since Spark starts
+	 * the server asynchronously, we just get a callback to this method, and the
+	 * only thing we can do to not lose data unknowingly is terminating the
+	 * process.
+	 */
+	private void logInitExceptionAndExit(Exception e) {
 		logger.error("Agent initialization failed", e);
+		System.exit(1);
```
DreierF: Not sure if this is the best way to handle this. I would have expected the agent to continue with the portscan from where it left if the initialization failed. I fear that this behaviour could lead to very hard to debug scenarios, but maybe I'm wrong. I'm not very familiar with micro-service setups ;-)

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
DreierF: Makes sense

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageAgent.java
DreierF: Sorry I missed that ;-)

karottenreibe:
Closing this as discussed with @agoeb . Reopen and update when this is actually needed.

Issue #68, opened 2019-05-09
/test/end should allow to pass test execution result


DreierF:

DreierF:
Will be available with 11.4.0

PR #69, opened 2019-05-10
Added option to supply TestExecution result via REST endpoint

DreierF:
- [ ] CHANGELOG.md updated

### DIFF
--- a/agent/src/main/java/com/teamscale/jacoco/agent/AgentOptions.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/AgentOptions.java
@@ -31,7 +31,6 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
-import java.util.Locale;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -207,7 +206,7 @@ public class AgentOptions {
 	}
 
 	private File getTempFile(final String prefix, final String extension) {
-		DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss.SSS", Locale.US);
+		DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss.SSS");
 		return new File(outputDirectory.toFile(),
 				prefix + "-" + dateFormat.format(new Date()) + "." + extension);
 	}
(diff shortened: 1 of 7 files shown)


DreierF:
Fixes #68

agent/src/main/java/com/teamscale/jacoco/agent/AgentOptions.java
```diff
@@ -199,21 +200,25 @@ public String createJacocoAgentOptions() {
...
-			return "sessionid=,destfile=" + new File(outputDirectory.toFile(),
-					"jacoco-" + dateFormat.format(new Date()) + ".exec").getAbsolutePath();
+			return "sessionid=,destfile=" + getTempFile("jacoco", "exec").getAbsolutePath();
 		} else {
 			return "output=none";
 		}
 	}
 
+	private File getTempFile(final String prefix, final String extension) {
+		DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss.SSS", Locale.US);
```
albertsteckermeier: Why do we hard-code `Locale.US` here?

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestExecutionWriter.java
```diff
@@ -0,0 +1,39 @@
...
+		}
+		FileReader fileReader = new FileReader(testExecutionFile);
+		List<TestExecution> testExecutions = GSON
+				.fromJson(fileReader, new TypeToken<List<TestExecution>>() {
+				}.getType());
+		fileReader.close();
+		testExecutions.add(testExecution);
+		FileWriter writer = new FileWriter(testExecutionFile);
+		GSON.toJson(testExecutions, writer);
+		writer.close();
```
albertsteckermeier: Why do we read the whole file and then write it again for adding a test execution. Can't we write the result just once for all test executions?

DreierF: The problem is that we will never know when we are done with the test execution.
We could only collect the test executions in memory and write them all to disk each time the list is updated.

agent/src/main/java/com/teamscale/jacoco/agent/AgentOptions.java
DreierF: Not sure why this ended up here. Removed it

PR #70, opened 2019-05-10
Added HTTP endpoints for non TIA mode (#61)

DreierF:
- [ ] CHANGELOG.md updated

### DIFF
--- a/agent/src/dist/logging/logback.console.xml
+++ b/agent/src/dist/logging/logback.console.xml
@@ -9,4 +9,7 @@
   <root level="INFO">
     <appender-ref ref="CONSOLE"/>
   </root>
+
+  <logger name="shadow.org.apache.spark" level="warn" />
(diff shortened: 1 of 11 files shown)


DreierF:
Closes #61

agent/src/dist/logging/logback.console.xml
```diff
@@ -9,4 +9,7 @@
   <root level="INFO">
     <appender-ref ref="CONSOLE"/>
   </root>
+
+  <logger name="shadow.org.apache.spark" level="warn" />
+  <logger name="shadow.org.eclipse.jetty" level="warn" />
```
albertsteckermeier: Is this leftover logging configuration change from debugging?

DreierF: No this was intentional, as spark and jetty are very verbose and spam the whole log which is not what we are interested in. We want to know if there are any problems, but not why which route was taken to serve the request etc.

PR #98, opened 2019-10-29
Fixes startup issues with WildFly 17+

DreierF:
Addresses issue #100

- [x] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/store/upload/teamscale/TeamscaleUploadStore.java
```diff
@@ -57,6 +47,14 @@ private boolean tryUploading(String xml) {
 		logger.debug("Uploading JaCoCo artifact to {}", teamscaleServer);
 
 		try {
+			// Cannot be executed in the constructor as this causes issues in WildFly server
+			// Internally this somehow interferes with the JBoss LogManager
```
karottenreibe: Please reference a GitHub issue here, so we can later trace this back to the cause. I think ATM there is no issue. Please create one, reference it in the pull request and describe what the problem was in more detail in it.

DreierF: Done

PR #101, opened 2019-11-25
Run all prioritized

DreierF:
Addresses issue #

- [ ] CHANGELOG.md updated

### DIFF
--- a/impacted-test-engine/src/main/java/com/teamscale/test_impacted/engine/executor/ImpactedTestsProvider.java
+++ b/impacted-test-engine/src/main/java/com/teamscale/test_impacted/engine/executor/ImpactedTestsProvider.java
@@ -28,11 +28,15 @@ public class ImpactedTestsProvider {
 
 	private final String partition;
 
-	public ImpactedTestsProvider(TeamscaleClient client, Long baseline, CommitDescriptor endCommit, String partition) {
+	private final boolean includeNonImpacted;
+
+	public ImpactedTestsProvider(TeamscaleClient client, Long baseline, CommitDescriptor endCommit, String partition,
+								 boolean includeNonImpacted) {
 		this.client = client;
 		this.baseline = baseline;
 		this.endCommit = endCommit;
 		this.partition = partition;
+		this.includeNonImpacted = includeNonImpacted;
 	}
 
 	/** Queries Teamscale for impacted tests. */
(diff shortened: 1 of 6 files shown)


scharfstein-dev:
Please check Teamscale Finding "unused import"

impacted-test-engine/src/main/java/com/teamscale/test_impacted/engine/executor/ImpactedTestsProvider.java
```diff
@@ -57,4 +62,24 @@ public ImpactedTestsProvider(TeamscaleClient client, Long baseline, CommitDescri
...
+		if (!this.includeNonImpacted) {
+			return true;
+		}
+		long returnedTests = testClusters.stream().mapToLong(g -> g.tests.size()).sum();
+		if (returnedTests == availableTestDetails.size()) {
+			return true;
+		} else {
+			LOGGER.error(
+					() -> "Retrieved " + returnedTests + " tests from Teamscale, but expected " + availableTestDetails
+							.size() + ".");
```
scharfstein-dev: Just wanted to make sure you knew that the following form would also work: `LOGGER.error("Retrieved {} tests from Teamscale, but expected {}.", returnedTests, availableTestDetails.size());`
If you prefer the explicit form, that's okay.

scharfstein-dev: Oh wait, I overlooked the lambda

scharfstein-dev: So yeah, with lambdas it looks like this: `LOGGER.error("Retrieved {} tests from Teamscale, but expected {}.", () -> returnedTests, () -> availableTestDetails.size());`
Actually I think you can just lose the lambda, since computing this is basically costing you nothing.

teamscale-client/src/main/java/com/teamscale/client/TeamscaleClient.java
```diff
@@ -44,12 +44,13 @@ public TeamscaleClient(String baseUrl, String user, String accessToken, String p
 	 */
 	public Response<List<PrioritizableTestCluster>> getImpactedTests(List<ClusteredTestDetails> testList, Long baseline,
 																	 CommitDescriptor endCommit,
-																	 String partition) throws IOException {
+																	 String partition,
+																	 boolean includeNonImpacted) throws IOException {
```
scharfstein-dev: ?

teamscale-client/src/main/java/com/teamscale/client/TeamscaleClient.java
```diff
@@ -64,14 +65,15 @@ public TeamscaleClient(String baseUrl, String user, String accessToken, String p
 	private Response<List<PrioritizableTestCluster>> getImpactedTestsOnce(List<ClusteredTestDetails> testList,
 																		  Long baseline,
 																		  CommitDescriptor endCommit,
-																		  String partition) throws IOException {
+																		  String partition,
+																		  boolean includeNonImpacted) throws IOException {
```
scharfstein-dev: What is happening with these empty lines? Is the formatter disabled?

impacted-test-engine/src/main/java/com/teamscale/test_impacted/engine/executor/ImpactedTestsProvider.java
DreierF: This is a `org.junit.platform.commons.logging.Logger`, which only has the supplier interface

teamscale-client/src/main/java/com/teamscale/client/TeamscaleClient.java
DreierF: You need to scroll to the right 😉
We use the default formatter of intellij here and it aligns the parameters with the opening parenthesis.

PR #108, opened 2020-02-19
35 decrease memory footprint

AnonymFx:
Addresses issue [TS-21140](https://jira.cqse.eu/browse/TS-21140)

- [x] CHANGELOG.md updated


karottenreibe:
a note for next time: the renaming `store` -> `uploader` makes this diff very cumbersome to review. Next time I'd prefer to separate this into a separate pull request that we can merge after the actual functional changes. That would make reviewing the functional changes for correctness much easier :-)

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/UploadStoreBase.java
```diff
@@ -94,38 +92,38 @@ protected boolean tryUpload(String xml) {
...
+				fillZipFile(zipOutputStream, coverageFile);
 			}
-			return byteArrayOutputStream.toByteArray();
+			return zipFile;
 		}
 	}
 
+	private File createZipFileObject(File coverageFile) {
+		Path parentFolder = coverageFile.getParentFile().toPath();
+		String filename = coverageFile.getName().substring(0, coverageFile.getName().length() - 4) + ".zip";
```
karottenreibe: please use `FileSystemUtils.getFilenameWithoutExtension` instead as that conveys what's happening here better

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/UploadStoreBase.java
```diff
@@ -94,38 +92,38 @@ protected boolean tryUpload(String xml) {
...
+		return new File(parentFolder.resolve(filename).toString());
```
karottenreibe: please use `toFile` instead of `new File(...toString())`

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/UploadStoreBase.java
```diff
@@ -94,38 +92,38 @@ protected boolean tryUpload(String xml) {
...
 	 */
-	private byte[] createZipFile(String xml) throws IOException {
-		try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
-			try (ZipOutputStream zipOutputStream = new ZipOutputStream(byteArrayOutputStream);
-				 OutputStreamWriter writer = new OutputStreamWriter(zipOutputStream, "UTF-8")) {
-				fillZipFile(zipOutputStream, writer, xml);
+	private File createZipFile(File coverageFile) throws IOException {
+		File zipFile = createZipFileObject(coverageFile);
+		try (FileOutputStream fileOutputStream = new FileOutputStream(zipFile)) {
+			try (ZipOutputStream zipOutputStream = new ZipOutputStream(fileOutputStream)) {
```
karottenreibe: since you're not catching exceptions here, you can join the two try-with-resources: `try(A a = ..., B b = ...) { }`

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/UploadStoreBase.java
```diff
@@ -53,33 +48,36 @@ public UploadStoreBase(TimestampedFileStore failureStore, HttpUrl uploadUrl, Lis
...
 			return false;
 		}
 
 		try {
-			Response<ResponseBody> response = uploadCoverageZip(zipFileBytes);
+			Response<ResponseBody> response = uploadCoverageZip(zipFile);
 			if (response.isSuccessful()) {
+				zipFile.delete();
 				return true;
 			}
```
karottenreibe: comment for the error we log later on (GitHub won't let me add comments in non-changed lines :-( ):
in case no error body is available, we currently log a newline and then nothing. This looks ugly in the logs. I'd prefer to log `<no server response>` or something like that in case the server does not return any body text to make it clear to the user what is happening

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/UploadStoreBase.java
```diff
@@ -94,38 +92,38 @@ protected boolean tryUpload(String xml) {
...
-	private void fillZipFile(ZipOutputStream zipOutputStream, OutputStreamWriter writer, String xml)
+	private void fillZipFile(ZipOutputStream zipOutputStream, File coverageFile)
 			throws IOException {
 		zipOutputStream.putNextEntry(new ZipEntry("coverage.xml"));
-		writer.write(xml);
-
-		// We flush the writer, but don't close it here, because closing the writer
-		// would also close the zipOutputStream, making further writes impossible.
-		writer.flush();
+		zipOutputStream.write(FileSystemUtils.readFileBinary(coverageFile));
```
karottenreibe: this again reads in the entire XML file as a byte array. Isn't that what we want to avoid at all costs, since the XMLs can be big? Maybe rather use `FileSystemUtils.copy` to copy the data over in chunks?

It's very easy to make this mistake. Should we create an abstraction `CoverageFile` that encapsulates and hides a private `Path` to a coverage file on disk and only offers a `copy` method to copy the file contents in this way into an output stream? That way, future maintainers are forced to think about this memory consumption issue and can't just accidentally read in the entire large file into memory.

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/azure/AzureFileStorageUploadStore.java
```diff
@@ -176,24 +176,25 @@ private String createFileName() {
...
-	private Response<ResponseBody> createAndFillFile(byte[] zipFilBytes, String fileName) throws UploadStoreException, IOException {
-		Response<ResponseBody> response = createFile(zipFilBytes, fileName);
+	private Response<ResponseBody> createAndFillFile(File zipFile,
+													 String fileName) throws UploaderException, IOException {
+		Response<ResponseBody> response = createFile(zipFile, fileName);
 		if (response.isSuccessful()) {
-			return fillFile(zipFilBytes, fileName);
+			return fillFile(zipFile, fileName);
 		}
 		logger.warn(String.format("Creation of file '%s' was unsuccessful.", fileName));
```
karottenreibe: why is this a warning and not an error? sounds like the upload fails in this case

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStore.java
```diff
@@ -1,99 +1,111 @@
...
+import java.util.stream.Stream;
 
 /**
- * Wraps an {@link IXmlStore} and an {@link ICachingXmlStore} in order to delay upload to the {@link IXmlStore} until a
- * {@link CommitDescriptor} is asynchronously made available. Until then, all XMLs are stored in the {@link
- * ICachingXmlStore}.
+ * Wraps an {@link IUploader} and in order to delay upload  until a {@link CommitDescriptor} is asynchronously made
+ * available.
  */
-public class DelayedCommitDescriptorStore implements IXmlStore {
```
karottenreibe: still reference to `store` in this class name

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStore.java
```diff
@@ -1,99 +1,111 @@
...
+public class DelayedCommitDescriptorStore implements IUploader {
 
 	private final Executor executor;
 	private final Logger logger = LoggingUtils.getLogger(this);
-	private final Function<CommitDescriptor, IXmlStore> wrappedStoreFactory;
-	private IXmlStore wrappedStore = null;
-	private final ICachingXmlStore cache;
+	private final Function<CommitDescriptor, IUploader> wrappedUploaderFactory;
+	private IUploader uploader = null;
```
karottenreibe: please rename to `wrappedUploader`

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStore.java
```diff
@@ -1,99 +1,111 @@
...
-		if (wrappedStore == null) {
-			wrappedStore = wrappedStoreFactory.apply(commit);
+		if (uploader == null) {
+			uploader = wrappedUploaderFactory.apply(commit);
 			logger.info("Commit to upload to has been found: {}. Uploading any cached XMLs now to {}", commit,
-					wrappedStore.describe());
+					uploader.describe());
 			executor.execute(this::uploadCachedXmls);
 		} else {
 			logger.error("Tried to set upload commit multiple times (old store: {}, new commit: {})." +
```
karottenreibe: still reference to `store` in log message

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStore.java
```diff
@@ -1,99 +1,111 @@
...
 	private void uploadCachedXmls() {
-		cache.streamCachedXmls().forEach(xml -> wrappedStore.store(xml));
-		cache.clear();
-		logger.debug("Finished upload of cached XMLs to {}", wrappedStore.describe());
+		try {
+			Stream<Path> xmlFilesStream = Files.list(cacheDir).filter(path -> {
+				String fileName = path.getFileName().toString();
+				return fileName.startsWith("jacoco-") && fileName.endsWith(".xml");
+			});
+			xmlFilesStream.forEach(file -> uploader.upload(file.toFile()));
```
karottenreibe: rename `file` to `path`?

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/teamscale/TeamscaleUploader.java
```diff
@@ -4,46 +4,43 @@
...
 	}
 
 	@Override
-	public void store(String xml) {
+	public void upload(File coverageFile) {
 		try (Benchmark benchmark = new Benchmark("Uploading report to Teamscale")) {
-			if (!tryUploading(xml)) {
-				logger.warn("Storing failed upload in {}", failureStore.getOutputDirectory());
-				failureStore.store(xml);
+			if (!tryUploading(coverageFile)) {
```
karottenreibe: invert condition for readability?

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/teamscale/TeamscaleUploader.java
```diff
@@ -4,46 +4,43 @@
...
+				logger.warn("Filed to upload coverage to Teamscale. Won't delete local file");
```
karottenreibe: typo: "Filed" -> "Failed"
also: please let the user know which file you won't delete and what they can do with it (upload manually)

agent/src/test/java/com/teamscale/jacoco/agent/TestBase.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Base class that should be used for all tests.
+ * Removes coverage directories created by the tests.
+ */
+public class TestBase {
```
karottenreibe: a lot of tests don't create coverage directories, so I don't like this solution as it makes it impossible to inherit from some other base class, should we ever need that. We had that in ConQAT times where everything needed to be a `CCSMTestCaseBase` which was cumbersome at times. So I'd like to avoid this if possible.

Could you please try to turn this into a JUnit 5 `@Rule` and use this only in the tests that actually create such directories? (I'm guessing: only integration/system tests).

Alternatively: change the code so the tests write to a temp directory (either the OS temp dir or a `test-tmp` directory that is .gitignore'd)

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorRetrievalTest.java
```diff
@@ -16,23 +18,25 @@
...
+		store.upload(coverageFile);
 		locator.searchJarFileForGitPropertiesAsync(new File(getClass().getResource("git-properties.jar").toURI()));
 		locatorExecutor.shutdown();
 		locatorExecutor.awaitTermination(5, TimeUnit.SECONDS);
 		storeExecutor.shutdown();
 		storeExecutor.awaitTermination(5, TimeUnit.SECONDS);
 
-		assertThat(cache.getXmls()).isEmpty();
-		assertThat(destination.getXmls()).containsExactly("xml1");
+		assertThat(Files.list(outputPath).anyMatch(path -> path.getFileName().equals(coverageFile.toPath())))
```
karottenreibe: is this assertion really correct? you're filtering for paths where the file name matches the entire path of the coverage file. I'd wager that is never true in any circumstances. shouldn't the `equals` just compare both paths (i.e. without the `getFilename`) or both file names (i.e. add a `getFileName to the right hand side)? can you please try and provoke a test failure to check that the assertion actually works as intended?

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStoreTest.java
```diff
@@ -1,56 +1,74 @@
...
-		store.store("xml1");
+		InMemoryUploader destination = new InMemoryUploader();
+		DelayedCommitDescriptorStore store = new DelayedCommitDescriptorStore(commit -> destination, outputPath);
 
-		assertThat(cache.getXmls()).containsExactly("xml1");
-		assertThat(destination.getXmls()).isEmpty();
+		store.upload(coverageFile);
+
+		assertThat(
+				Files.list(outputPath).anyMatch(path -> path.getFileName().equals(Paths.get(coverageFile.getName()))))
```
karottenreibe: if you're comparing file names, I'd prefer `path.getFileName().toString().equals(coverageFile.getName())` that is easier to understand for me

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStoreTest.java
```diff
@@ -1,56 +1,74 @@
...
+				.isTrue();
```
karottenreibe: please convert the stream to a list and use proper list assertions, as these give nicer error masseges than `isTrue`

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorRetrievalTest.java
```diff
@@ -16,23 +18,25 @@
...
+				.isFalse();
```
karottenreibe: please convert the stream to a list and use proper list assertions, as these give nicer error masseges than `isFalse`

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorRetrievalTest.java
```diff
@@ -16,23 +18,25 @@
...
+		assertThat(destination.getUploadedFiles().contains(coverageFile)).isTrue();
```
karottenreibe: please use proper list assertions, as these give nicer error masseges than `isTrue`

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStoreTest.java
```diff
@@ -1,56 +1,74 @@
...
+		assertThat(destination.getUploadedFiles().contains(coverageFile)).isFalse();
```
karottenreibe: please use proper list assertions, as these give nicer error masseges than `isFalse`

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStoreTest.java
```diff
@@ -1,56 +1,74 @@
...
 
 		store.setCommitAndTriggerAsynchronousUpload(new CommitDescriptor("branch", 1234));
-		store.store("xml1");
+		store.upload(coverageFile);
 
-		assertThat(cache.getXmls()).isEmpty();
-		assertThat(destination.getXmls()).containsExactly("xml1");
+		assertThat(
+				Files.list(outputPath).anyMatch(path -> path.getFileName().equals(Paths.get(coverageFile.getName()))))
+				.isFalse();
```
karottenreibe: again: please don't use `isFalse`/`isTrue` if possible

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStoreTest.java
```diff
@@ -1,56 +1,74 @@
...
+		DelayedCommitDescriptorStore store = new DelayedCommitDescriptorStore(commit -> destination, outputPath);
 
 		store.setCommitAndTriggerAsynchronousUpload(new CommitDescriptor("branch", 1234));
-		store.store("xml1");
+		store.upload(coverageFile);
 
-		assertThat(cache.getXmls()).isEmpty();
-		assertThat(destination.getXmls()).containsExactly("xml1");
+		assertThat(
+				Files.list(outputPath).anyMatch(path -> path.getFileName().equals(Paths.get(coverageFile.getName()))))
```
karottenreibe: see my comment above about comparing file names

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorStoreTest.java
```diff
@@ -1,56 +1,74 @@
...
 
-		store.store("xml1");
+		store.upload(coverageFile);
 		store.setCommitAndTriggerAsynchronousUpload(new CommitDescriptor("branch", 1234));
 		executor.shutdown();
 		executor.awaitTermination(5, TimeUnit.SECONDS);
 
-		assertThat(cache.getXmls()).isEmpty();
-		assertThat(destination.getXmls()).containsExactly("xml1");
+		assertThat(
```
karottenreibe: same comments as in the other test cases

report-generator/src/main/java/com/teamscale/report/jacoco/JaCoCoXmlReportGenerator.java
```diff
@@ -53,11 +52,36 @@ public JaCoCoXmlReportGenerator(List<File> codeDirectoriesOrArchives,
...
 		convertToReport(output, dump);
-		return output.toString(FileSystemUtils.UTF8_ENCODING);
+		return outputFile;
+	}
+
+	private File createOutputFile(String filePath) {
+		File outputFile = new File(filePath);
+
+		try {
+			boolean fileCreated = outputFile.createNewFile();
```
karottenreibe: please use java NIO where possible as it uses exceptions instead of boolean return codes, so we are forced to check the error cases and can't forget accidentally. In this case: `Files.createFile`.  also, in this case, you don't log anything which is bad for the user since they don't know that anything went wrong but don't see any coverage either. So they have no way of knowing what the problem is and thus no way of fixing it themselves

report-generator/src/main/java/com/teamscale/report/jacoco/JaCoCoXmlReportGenerator.java
```diff
@@ -53,11 +52,36 @@ public JaCoCoXmlReportGenerator(List<File> codeDirectoriesOrArchives,
...
+			if (!fileCreated) {
+				return null;
+			}
+		} catch (IOException e) {
+			logger.error("Could not create file " + outputFile.getAbsolutePath() + ". ");
```
karottenreibe: please let the user know what that means and what they can do about it. Otherwise this error message is more confusing than helpful. They don't know what that file is supposed to do (ideally: they don't even know we cache stuff on disk).
Also: please **always** log the exception object. It almost always contains useful details and helps debug the root cause (e.g. disk full, permission problems, ...)
I'd prefer something like `Could not create temporary coverage file {file}. This is used to cache the coverage file on disk before uploading it to its final destination. This coverage is lost. Please fix the underlying issue to avoid losing coverage.`
or similar

agent/src/main/java/com/teamscale/jacoco/agent/store/IUploader.java
```diff
@@ -0,0 +1,14 @@
+package com.teamscale.jacoco.agent.store;
+
+import java.io.File;
+
+/** Uploads coverage reports. */
+public interface IUploader {
+
+	/** Uploads the given coverage file. */
```
karottenreibe: please document that the uploader is responsible for deleting the coverage file in case the upload succeeds and that it should be left on disk if not (plus logged to the user). This is now a new requirement for all uploaders that wasn't previously there.

karottenreibe:
general comment: please rename the `store` package to `upload` for consistency

agent/src/main/java/com/teamscale/jacoco/agent/store/NoopUploader.java
```diff
@@ -0,0 +1,18 @@
+package com.teamscale.jacoco.agent.store;
+
+import java.io.File;
+
+/**
+ * Dummy uploader that does not provide any functionality. Can be used instead of null to mimic no upload behaviour.
+ */
+public class NoopUploader implements IUploader {
```
karottenreibe: noop is not really what it does though: it's an uploader that saves the xml files to disk. Thus, I'd prefer to call it `LocalDirectoryUploader` or something like that. It's an uploader that always succeeds and thus never deletes the cached files. This should be documented

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/UploadStoreBase.java
```diff
@@ -94,38 +92,38 @@ protected boolean tryUpload(String xml) {
...
 
 		for (Path additionalFile : additionalMetaDataFiles) {
 			zipOutputStream.putNextEntry(new ZipEntry(additionalFile.getFileName().toString()));
```
karottenreibe: the next line also loads arbitrary files into memory that might be huge. we should also stream this instead of caching them in memory

karottenreibe:
Please also update the CHANGELOG

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -33,22 +35,27 @@
...
+	public Agent(AgentOptions options,
+				 Instrumentation instrumentation) throws IllegalStateException, UploaderException {
 		super(options);
 
-		store = options.createStore(instrumentation);
-		logger.info("Storage method: {}", store.describe());
+		uploader = options.createUploader(instrumentation);
+		logger.info("Storage method: {}", uploader.describe());
+
+		this.outputDirectory = options.getOutputDirectory();
```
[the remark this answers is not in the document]
AnonymFx: We need the output directory in the method `dumpReportUnsafe`, so we need to introduce a field for either `outputDirectory` or for the agent options (currently we use those only in the constructor). Do you think there is a benefit of referencing the agent options over only the output directory?

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -142,14 +149,19 @@ private void dumpReportUnsafe() {
 			return;
 		}
 
-		String xml;
+		File coverageFile;
+		long currentTime = System.currentTimeMillis();
+		Path outputFile = this.outputDirectory.resolve("jacoco-" + currentTime + ".xml");
 		try (Benchmark benchmark = new Benchmark("Generating the XML report")) {
-			xml = generator.convert(dump);
+			coverageFile = generator.convert(dump, outputFile.toString());
```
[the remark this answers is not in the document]
AnonymFx: Good point :D

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -142,14 +149,19 @@ private void dumpReportUnsafe() {
...
 		} catch (IOException e) {
 			logger.error("Converting binary dump to XML failed", e);
 			return;
 		}
 
-		store.store(xml);
+		if (coverageFile != null) {
```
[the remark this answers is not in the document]
AnonymFx: Could have happened when `ouputFile.createNewFile()` in line 76 of class `JaCoCoXmlReportGenerator` returned false. I read the documentation again, this would happen if the file already existed. So instead of returning null, I changed the code to clear the file contents instead. => we don't need this null check anymore :)

AnonymFx:
Renamed the remaining store classes as well, since we change all references in the code anyways. Sorry if that makes the review more complicated again 😬

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
```diff
@@ -312,44 +317,34 @@ public AgentBase createAgent(
...
-				return createDelayedTeamscaleStore(instrumentation, fileStore);
+				return createDelayedTeamscaleStore(instrumentation);
 			}
-			return new TeamscaleUploadStore(fileStore, teamscaleServer);
+			return new TeamscaleUploader(teamscaleServer);
 		}
 
 		if (azureFileStorageConfig.hasAllRequiredFieldsSet()) {
-			return new AzureFileStorageUploadStore(fileStore, azureFileStorageConfig,
+			return new AzureFileStorageUploadStore(azureFileStorageConfig,
```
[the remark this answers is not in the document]
AnonymFx: That's actually fine in this case, the "storage" in "azureFileStorage" does not reference our store/upload concept, but rather the file store on azure.

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/UploadStoreBase.java
AnonymFx: I like the idea. Unfortunately, we need a `java.io.Filo` to create the request body for the Teamscale upload, for example. So we need to expose the File anyways. I'll create a `CoverageFile` class anyway, as it makes it at least a little bit harder to do the wrong thing and we can advise against using the getter for the `java.io.File` in the documentation.

agent/src/main/java/com/teamscale/jacoco/agent/store/upload/azure/AzureFileStorageUploadStore.java
AnonymFx: No clue, didn't write that line 🤷‍♂️Will change it to error, though.

agent/src/test/java/com/teamscale/jacoco/agent/TestBase.java
AnonymFx: Removed the class and added the `@Afterall` to the two classes causing problems. Probably not the most elegant solution, but I'd like to get the ticket resolved rather quick, so that's my compromise. If you think that we want to do it a different way (with JUnit5 Extensions for example, @Rule doesn't exist anymore) please create a new maintenance ticket for it.

agent/src/test/java/com/teamscale/jacoco/agent/store/upload/delay/DelayedCommitDescriptorRetrievalTest.java
AnonymFx: yeah, that's a mistake, removed the `getFileName` call.

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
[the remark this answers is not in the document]
karottenreibe: no, makes sense :-) didn't see that!

agent/src/main/java/com/teamscale/jacoco/agent/AgentBase.java
```diff
@@ -17,7 +17,7 @@
  * Base class for agent implementations. Handles logger shutdown, store creation and instantiation of the {@link
  * JacocoRuntimeController}.
  * <p>
- * Subclasses must handle dumping into the store.
+ * Subclasses must handle dumping onto disk.
```
karottenreibe: and also uploading via the configured uploader, right?

agent/src/main/java/com/teamscale/jacoco/agent/upload/LocalDiskUploader.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+public class LocalDiskUploader implements IUploader {
+	@Override
+	public void upload(CoverageFile coverageFile) {
+		// Don't delete the file here. We want to store the file permanently on disk in case no uploader is configured.
+	}
+
+	@Override
+	public String describe() {
+		return "configured output directory on the local disk without upload";
```
karottenreibe: please remove "without upload". That is confusing for me

agent/src/main/java/com/teamscale/jacoco/agent/upload/teamscale/TeamscaleUploader.java
```diff
@@ -34,18 +30,21 @@ public TeamscaleUploader(TeamscaleServer teamscaleServer) {
...
-			if (!tryUploading(coverageFile)) {
-				logger.warn("Filed to upload coverage to Teamscale. Won't delete local file");
-			} else {
+			if (tryUploading(coverageFile)) {
 				coverageFile.delete();
+			} else {
+				logger.warn("Failed to upload coverage to Teamscale. Won't delete local file {}",
+						coverageFile.getAbsolutePath());
 			}
+		} catch (IOException e) {
```
karottenreibe: if this is truly just for the `delete()` call, please move this around that. otherwise, something else might (in the future) raise that exception and we would log a nonsensical message then
also, please don't swallow the exception, log it
also, please tell the user what that means, i.e: "The upload was successful, but the file will be left on disk. You can delete it yourself anytime - it is no longer needed" or something similar

agent/src/main/java/com/teamscale/jacoco/agent/upload/teamscale/TeamscaleUploader.java
```diff
@@ -34,18 +30,21 @@ public TeamscaleUploader(TeamscaleServer teamscaleServer) {
...
-	public void upload(File coverageFile) {
+	public void upload(CoverageFile coverageFile) {
 		try (Benchmark benchmark = new Benchmark("Uploading report to Teamscale")) {
-			if (!tryUploading(coverageFile)) {
-				logger.warn("Filed to upload coverage to Teamscale. Won't delete local file");
-			} else {
+			if (tryUploading(coverageFile)) {
 				coverageFile.delete();
+			} else {
+				logger.warn("Failed to upload coverage to Teamscale. Won't delete local file {}",
```
karottenreibe: ..."so you can upload it yourself manually"

agent/src/main/java/com/teamscale/jacoco/agent/upload/teamscale/TeamscaleUploader.java
```diff
@@ -63,7 +62,7 @@ private boolean tryUploading(File xml) {
 					teamscaleServer.partition,
 					EReportFormat.JACOCO,
 					teamscaleServer.message,
-					RequestBody.create(MultipartBody.FORM, xml)
+					RequestBody.create(MultipartBody.FORM, new File(coverageFile.getAbsolutePath()))
```
karottenreibe: I'd prefer a method `CoverageFile.createFormRequestBody()` so other people don't have to think about how to properly do this without wasting memory

report-generator/src/main/java/com/teamscale/report/jacoco/CoverageFile.java
```diff
@@ -0,0 +1,85 @@
...
+	 * Get the filename of the coverage file on disk without its extension
+	 */
+	public String getNameWithoutExtension() {
+		return FileSystemUtils.getFilenameWithoutExtension(coverageFile);
+	}
+
+	/**
+	 * Get the {@link java.io.File} representing the parent direcotry of the coverage file
+	 */
+	public Path getParentDirectoryPath() {
```
karottenreibe: this is only used in tests, where we already know the actual underlying file. Please remove from this class and inline this in the test cases instead

agent/src/test/java/com/teamscale/jacoco/agent/upload/delay/DelayedCommitDescriptorRetrievalTest.java
```diff
@@ -18,11 +20,11 @@
 	@Test
 	public void locatorShouldTriggerUploadOfCachedXmls() throws Exception {
 		ExecutorService storeExecutor = Executors.newSingleThreadExecutor();
-		File coverageFile = File.createTempFile("jacoco-", ".xml");
-		Path outputPath = coverageFile.getParentFile().toPath();
+		CoverageFile coverageFile = new CoverageFile(File.createTempFile("jacoco-", ".xml"));
```
karottenreibe: I'd prefer you create a temp *dir*, then create a file with a fixed name inside, as you're later iterating over the temp dir which may contain leftovers from other tests

karottenreibe: same for the other tests where you use `getParentDirectoryPath`

report-generator/src/main/java/com/teamscale/report/jacoco/CoverageFile.java
```diff
@@ -0,0 +1,85 @@
...
+		return coverageFile.getParentFile().toPath();
+	}
+
+	/**
+	 * Get the absolute path of the underlying {@link java.io.File}. Do only use it for logging purposes and especially
+	 * don't use it to read the entire file into memory.
```
karottenreibe: it's also used in tests, but maybe those references can be removed.
for logging, we could give this a toString() method and just call `logger.info("bla {} blu", coverageFile)`
then you could remove this and the caveats entirely

report-generator/src/main/java/com/teamscale/report/jacoco/JaCoCoXmlReportGenerator.java
```diff
@@ -58,27 +61,26 @@ public JaCoCoXmlReportGenerator(List<File> codeDirectoriesOrArchives,
...
-				return null;
-			}
+			Files.createFile(outputFile.toPath());
+		} catch (FileAlreadyExistsException e) {
+			// Clear file contents
+			new FileOutputStream(outputFile).close();
 		} catch (IOException e) {
-			logger.error("Could not create file " + outputFile.getAbsolutePath() + ". ");
-			return null;
+			// Add additional info to the IOException
```
karottenreibe: is this a TODO comment for you? if not, please either remove or explain as it's confusing for me what you mean with this

report-generator/src/main/java/com/teamscale/report/jacoco/JaCoCoXmlReportGenerator.java
```diff
@@ -58,27 +61,26 @@ public JaCoCoXmlReportGenerator(List<File> codeDirectoriesOrArchives,
...
-		}
 		FileOutputStream output = new FileOutputStream(outputFile);
 		convertToReport(output, dump);
-		return outputFile;
+		return new CoverageFile(outputFile);
 	}
 
-	private File createOutputFile(String filePath) {
-		File outputFile = new File(filePath);
+	private File createOutputFile(Path filePath) throws IOException {
```
karottenreibe: I'm not sure we need this. if writing to the file fails, we already get an IOException telling us the cause (e.g. missing permissions). I'm not sure creating the file before writing to it actually has any benefit. remove this method?

AnonymFx: As you argued for adapting the Message in the IOException: It might make it clearer to the user why we are even writing to disk. So I'd prefer to keep this method.

AnonymFx:
> see my comments.
> 
> as this is a very large change, please test at least the two most important upload variants (Teamscale, local disk) manually to make sure they still work as expected

Teamscale and local work as expected. Azure and Http should as well, at least it works to the point until zipping. And after zipping everything stayed the same.

report-generator/src/main/java/com/teamscale/report/jacoco/JaCoCoXmlReportGenerator.java
AnonymFx: I don't understand the 4th point. If we are still calling `createReport` with the `OutputStream` we can't handle the IOException from creating the coverageFile in it, because the call to create it (probably in `getOutputStream`) would already be in the method call not in the body.

I moved the exception to `getOutputStream` now and removed all code that handled creating the file specifically, as calling the constructor of `FileOutputStream` already creates the file if it does not exist. Please check if that's fine for you :)

AnonymFx:
I'm the one having to thank you for all the great feedback ;)

PR #125, opened 2020-03-26
Agent library test

karottenreibe:
Addresses issue  TS-22443

- [x] Agent's README.md updated in case of user-visible changes - not needed
- [x] CHANGELOG.md updated - not needed
- [ ] copy TS arch to master


karottenreibe:
My idea is that we can test e.g. the normal agent in a similar manner easily to have some E2E tests to make sure important functionality doesn't break completely

PR #135, opened 2020-05-04
CR#22849 Fixed duplicate tests

DreierF:
Addresses issue #

- [x] CHANGELOG.md updated


report-generator/src/main/java/com/teamscale/report/testwise/model/factory/TestInfoFactory.java
```diff
@@ -65,11 +65,20 @@ public TestInfo createFor(TestCoverageBuilder testCoverageBuilder) {
...
-			if (uniformPathsWithCoverage.contains(testDetails.uniformPath)) {
+			if (!uniformPathsWithCoverage.contains(testDetails.uniformPath)) {
 				TestInfoBuilder testInfo = new TestInfoBuilder(testDetails.uniformPath);
 				testInfo.setDetails(testDetails);
 				testInfo.setExecution(testExecutionsMap.get(testDetails.uniformPath));
 				results.add(testInfo.build());
+				uniformPathsWithCoverage.add(testDetails.uniformPath);
+			}
+		}
+		for (TestExecution testExecution : testExecutionsMap.values()) {
```
AnonymFx, line 76: Why do we need this second loop? Shouldn't all entries in the `testExecutionsMap` have coverage and are thus not relevant for the test infos **without** coverage?

agent/src/test/java/com/teamscale/jacoco/agent/convert/ConverterTest.java
```diff
@@ -33,4 +36,39 @@ public void testSmokeTest(@TempDir File tempDir) throws Exception {
...
+				contains(
+						"\"uniformPath\": \"[engine:junit-vintage]/[runner:org.conqat.lib.cqddl.CQDDLTest]/[test:testFunctions(org.conqat.lib.cqddl.CQDDLTest)]\"")
+				.contains(
+						"\"uniformPath\": \"[engine:junit-vintage]/[runner:org.conqat.lib.cqddl.CQDDLTest]/[test:testDirectObjectInsertion(org.conqat.lib.cqddl.CQDDLTest)]\"")
+				.contains("\"uniformPath\": \"[engine:junit-vintage]/[runner:org.conqat.lib.cqddl.CQDDLTest]/[test:testKeyAbbreviations(org.conqat.lib.cqddl.CQDDLTest)]\"")
+				.contains("\"uniformPath\": \"[engine:junit-vintage]/[runner:org.conqat.lib.cqddl.CQDDLTest]/[test:testKeyAbbreviations(org.conqat.lib.cqddl.CQDDLTest)]\"")
+				.contains("\"result\": \"PASSED\"").contains("\"duration\": 1234").contains("\"coveredLines\": \"33,46-47");
+	}
+
+	private void copyResourceTo(File inputDir, String name) throws URISyntaxException, IOException {
```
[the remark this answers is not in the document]
DreierF: Sure

report-generator/src/main/java/com/teamscale/report/testwise/model/factory/TestInfoFactory.java
DreierF: You are right! I just thought this would make it more robust if you do not provide all relevant artifacts, but testwise coverage with executed tests and no coverage is probably always a misconfiguration. Instead of adding the info I now print an error for those cases.

report-generator/src/main/java/com/teamscale/report/testwise/model/factory/TestInfoFactory.java
```diff
@@ -65,11 +65,18 @@ public TestInfo createFor(TestCoverageBuilder testCoverageBuilder) {
...
 				TestInfoBuilder testInfo = new TestInfoBuilder(testDetails.uniformPath);
 				testInfo.setDetails(testDetails);
 				testInfo.setExecution(testExecutionsMap.get(testDetails.uniformPath));
 				results.add(testInfo.build());
+				processedTestUniformPaths.add(testDetails.uniformPath);
+			}
+		}
+		for (TestExecution testExecution : testExecutionsMap.values()) {
+			if (!processedTestUniformPaths.contains(testExecution.getUniformPath())) {
+				System.err.println("No coverage found found for " + testExecution.getUniformPath());
```
AnonymFx, line 78: Maybe we can add some more details, why we'd expect coverage here and what to do about it, e.g.
`"Test ${testExecution.getUniformPath()} was executed but no coverage file was found. Please make sure that xyz/do xyz"`
I'm not sure what the user might be able to do to fix this problem. If you don't either, we could just say "Please make sure that you configured the agent correctly." or something similar.

PR #150, opened 2020-07-10
CR#0 Clarify meaning of "Filtering" in debug log

### LINKED ISSUE #149: Debug log talks about filtering, but it is include whether include or exclude is meant

sewe:
Closes #149


karottenreibe:
FYI @sewe the process for the Agent is the same as for Teamscale: small fixes like this can be committed CR#0 to master. Larger ones require a Jira Ticket :-) Sorry for not stating this clearly in Slack (i.e. that you can just CR#0 this to master)

PR #156, opened 2020-08-11
Cr/24034 override message via http post

AnonymFx:
Addresses issue [TS-24034](https://jira.cqse.eu/browse/TS-24034)

- [x] CHANGELOG.md updated
- [x] Present new features in [N&N](https://wiki.cqse.eu/pages/viewpage.action?pageId=689566) ([Scheduled for Aug 28](https://wiki.cqse.eu/pages/viewpage.action?pageId=37224751))
- [x] ~~[TGA Tutorial](https://docs.teamscale.com/tutorial/setting-up-tga-java) updated~~ Not a breaking change
- [x] ~~[TIA Tutorial](https://docs.teamscale.com/tutorial/tia-java/) updated~~ Not a breaking change


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ We use [semantic versioning](http://semver.org/):
 - [breaking change] removed options `coverage-via-http` and `teamscale-testwise-upload`. Use `tia-mode=http` or `tia-mode=teamscale-upload` instead.
 - [breaking change] _tia-client_: changed Java API
 - [feature] always exclude common libraries from profiling to shrink coverage files
+- [feature] enable changing and reading the commit message via http
```
sewe, line 11: "via http" is a bit imprecise. (HTTPS, too?) Better "via a new `/message` REST endpoint"

agent/README.md
```diff
@@ -150,6 +150,8 @@ echo `git rev-parse --abbrev-ref HEAD`:`git --no-pager log -n1 --format="%ct000"
     - `[POST] /partition/{partitionName}` Sets the name of the partition name that should be used for all followup 
       report dumps (see `teamscale-partition`). For reports that are not directly sent to Teamscale the generated report 
       will contain the partition name as session ID.
+    - `[GET] /message` Returns the name of the currently configured commit message.
+    - `[POST] /message/{newMessage}` Sets the commit message that should be used for all followup report dumps (see `teamscale-message`).
```
sewe, line 154: From a REST design standpoint, this is odd. Yes, `POST` has "anything you want semantics", but having the message as part of the UIR is weird, as it is not clear that `POST /message/New` overwrites a previous `POST /message/Previous`. It's all very implicit.

If I had to design the API, I would have done `GET /message` and `PUT /message` (with the new message in the body). Likewise for `/partition`. Here, the overwriting the previous message is even backed into the HTTP specs.

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -110,6 +116,23 @@ private String handleSetPartition(Request request, Response response) {
 		return "";
 	}
 
+	/** Handles setting the partition name. */
+	private String handleSetMessage(Request request, Response response) {
+		String message = request.params(MESSAGE_PARAMETER);
+		if (message == null || message.isEmpty()) {
+			logger.error("Message missing in " + request.url() + "! Expected /message/Some%20message.");
+
+			response.status(400);
```
sewe, line 125: Here and above, `HttpServletResponse.SC_BAD_REQUEST` would be clearer.

agent/src/main/java/com/teamscale/jacoco/agent/AgentBase.java
```diff
@@ -101,15 +101,20 @@ public static void premain(String options, Instrumentation instrumentation) thro
...
+			stopServer();
 			prepareShutdown();
 			logger.info("CQSE JaCoCo agent successfully shut down.");
 			loggingResources.close();
 		}));
 	}
 
+	/** Stop the http server if it's running */
+	void stopServer() {
+		if (options.getHttpServerPort() != null) {
```
sewe, line 113: This check looks odd, as it does not exactly scream "it is running". It just says that an option has been set.

As `Service.stop` already checks whether the service has been initialized (if no, it's a no-op), you can call it unconditionally. After all, it is only initialized in line 48 when the option is non-null.

agent/src/test/java/com/teamscale/jacoco/agent/AgentHttpServerTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class AgentHttpServerTest {
+
+	private Agent agent;
+	private final String defaultCommitMessage = "Some Message";
+
+	/** Starts the http server to control the agent */
+	@BeforeEach
+	public void setup() throws AgentOptionParseException, UploaderException {
+		AgentOptions options = AgentOptionsParser.parse("" +
```
sewe, line 28: As this does not test the `AgentOptionsParser`, I would prefer if it simply constructed the `AgentOptions` directly, perhaps with the help of some `TestAgentOptionsBuilder` (using the Test Data Builder pattern). See for example `TestNamedServerDetailsBuilder` in the Teamscale codebase.

Then, only explicitly set the options in the test that are actually relevant to understand this particular test case.

agent/README.md
AnonymFx: Form that perspective, I agree. But I'm still not sure if we want to change that from a compatibility standpoint.
@karottenreibe @DreierF What do you think?

agent/src/main/java/com/teamscale/jacoco/agent/AgentBase.java
AnonymFx: Tried removing the check and got a Nullpointer on `routes.clear()` in the `Service.stop()` method "in Tread-5". The JVM didn't crash, but this should also not appear in the logs, so I'd prefer keeping the check.

agent/README.md
DreierF: I agree from the conceptual standpoint. I guess this was done back then because it is easier to deal with in scripts etc. @goeb might have input on whether we can change that without affecting customers.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ We use [semantic versioning](http://semver.org/):
 - [breaking change] removed options `coverage-via-http` and `teamscale-testwise-upload`. Use `tia-mode=http` or `tia-mode=teamscale-upload` instead.
 - [breaking change] _tia-client_: changed Java API
 - [feature] always exclude common libraries from profiling to shrink coverage files
+- [feature] enable changing and reading the commit message via a new `/message` REST endpoint
+- [breaking change] changing the partition is now a PUT request with the new partition in the request body (formerly POST with the new partition in the URL)
```
sewe, line 12: The whole `with` description is a bit wordy. I think `/partition/{partitionName}` and `/partition` are more effective at communicating here.

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -71,42 +69,64 @@ public Agent(AgentOptions options,
...
 		controller.reset();
-		response.status(204);
+		response.status(HttpServletResponse.SC_NO_CONTENT);
 		return "";
 	}
 
 	/** Handles setting the partition name. */
 	private String handleSetPartition(Request request, Response response) {
-		String partition = request.params(PARTITION_PARAMETER);
+		String partition = request.body();
```
sewe, line 98: Is the `charset=UTF-8` requirement actually necessary or does Spark does any magic on `Request.body()`? In other words, Spark needs to pick some encoding to convert from bytes to `String` and I wonder how it does this here. Can you please check.

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -71,42 +69,64 @@ public Agent(AgentOptions options,
...
+
+	/** Handles setting the partition name. */
+	private String handleSetMessage(Request request, Response response) {
+		String message = request.body();
+		if (message == null || message.isEmpty()) {
+			logger.error("The new message is missing in the request body! Please add it as plain text " +
+					"(content type \"text/plain; charset=UTF-8\"");
+
+			response.status(HttpServletResponse.SC_BAD_REQUEST);
+			return "Message is missing!";
```
sewe, line 123: Would the error string not also make a got return value? Why be more verbose/helpful on only one channel?

agent/src/test/java/com/teamscale/jacoco/agent/AgentHttpServerTest.java
```diff
@@ -0,0 +1,104 @@
...
+	/** Starts the http server to control the agent */
+	@BeforeEach
+	public void setup() throws AgentOptionParseException, UploaderException {
+		AgentOptions options = new TestAgentOptionsBuilder()
+				.withHttpServerPort(httpServerPort)
+				.withTeamscaleMessage(defaultCommitMessage)
+				.withTeamscalePartition(defaultPartition)
+				.create();
+
+		// Starts the HTTP server for the agent
```
sewe, line 37: Captain Obvious to the rescue!

agent/src/test/java/com/teamscale/jacoco/agent/AgentHttpServerTest.java
```diff
@@ -0,0 +1,104 @@
...
+public class AgentHttpServerTest {
+
+	private Agent agent;
+	private final Integer httpServerPort = 8081;
+	private final String defaultCommitMessage = "Some Message";
+	private final String defaultPartition = "Some Partition";
+
+	/** Starts the http server to control the agent */
+	@BeforeEach
+	public void setup() throws AgentOptionParseException, UploaderException {
```
sewe, line 30: General advice for tests. Don't bother with "precise" `throws` clauses. Just use `throws Exception` anywhere, as there is no-one handling the different types of exception specially. Just the JUnit runner doing the same for all exceptions: failing the tests.

agent/src/test/java/com/teamscale/jacoco/agent/AgentHttpServerTest.java
```diff
@@ -0,0 +1,104 @@
...
+	/** Test reading the partition */
+	@Test
+	public void testGettingPartition() throws IOException {
+		String receivedPartition = readFromEndpoint("partition");
+
+		assertThat(receivedPartition).isEqualTo(defaultPartition);
+	}
+
+
+	private void overrideEndpoint(String endpointPath, String newValue) throws IOException {
```
sewe, line 86: I find these method names far to abstract/removed from what they actually do. Why not `putText("/message", "some message")` and `getText("/message")`?

agent/src/test/java/com/teamscale/jacoco/agent/AgentHttpServerTest.java
```diff
@@ -0,0 +1,104 @@
...
+		OkHttpClient client = new OkHttpClient();
+		MediaType textPlainMediaType = MediaType.parse("text/plain; charset=utf-8");
+		Request request = new Request.Builder()
+				.url("http://localhost:" + httpServerPort + "/" + endpointPath)
```
sewe, line 90: I would put `http://localhost:" + httpServerPort` into a field of type `URI` and than use `baseUri.resolve(endpointPath)` here. String operations on URIs are an Anti-pattern IMHO, as we have a better API with `URI.resolve`.

agent/src/test/java/com/teamscale/jacoco/agent/AgentHttpServerTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+	private Agent agent;
+	private final Integer httpServerPort = 8081;
+	private final String defaultCommitMessage = "Some Message";
+	private final String defaultPartition = "Some Partition";
+
+	/** Starts the http server to control the agent */
+	@BeforeEach
+	public void setup() throws AgentOptionParseException, UploaderException {
+		AgentOptions options = new TestAgentOptionsBuilder()
```
sewe, line 31: I generally prefer a `static` `someAgentOptions` factory method, that makes sure that all fields have valid values (so I can immediately call `create` if I don't want to override anything). In this case in particular, I get the feeling that simply the default options are what we want in the test; any customization only suggests that the custom values are somehow important for the test.

sewe (changes requested):
A couple of comments inline.

agent/src/test/java/com/teamscale/jacoco/agent/AgentHttpServerTest.java
AnonymFx: They are, assuming that we won't only be using the builder for the `AgentHttpServerTest` in the future.
* `withHttpServerPort` is required to be set in order to start the http server
* `withTeamscaleMessage` is relevant for the two message tests
* `withTeamscalePartition` is relevant for the two partition tests
Since we don't need to set any other options for the agent to work, I saw no sense in providing a static factory method.

AnonymFx: Essentially, the current implementation is a compromise between implementing the builder to support all agent options (which we don't need yet) and implementing it in a way that it only fits the `AgentHttpServerTest` (i.e. setting the required options as default options in a static factory method). This way, we can easily extend it in the future.

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
AnonymFx: It reads the charset from the request, so UTF-8 should not be required. Removed it from the message.

sewe (approved):
Looks good to me now. Thanks for the rework.

PR #158, opened 2020-09-04
Saner default message

karottenreibe:
Addresses issue https://jira.cqse.eu/browse/TS-24326

- [x] CHANGELOG.md updated

### DIFF
--- a/teamscale-gradle-plugin/src/test/kotlin/com/teamscale/TeamscalePluginTest.kt
+++ b/teamscale-gradle-plugin/src/test/kotlin/com/teamscale/TeamscalePluginTest.kt
@@ -135,7 +135,7 @@ class TeamscalePluginTest {
             .withProjectDir(temporaryFolder.root)
             .withPluginClasspath()
             .withArguments(runnerArgs)
-            .withGradleVersion("4.6")
+            .withGradleVersion("6.5")
 
         if (DEBUG_PLUGIN) {
             runner.withDebug(true)
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -9,6 +9,22 @@
       <option name="JD_DO_NOT_WRAP_ONE_LINE_COMMENTS" value="true" />
     </JavaCodeStyleSettings>
     <JetCodeStyleSettings>
+      <option name="PACKAGES_TO_USE_STAR_IMPORTS">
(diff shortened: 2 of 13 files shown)


teamscale-gradle-plugin/src/test/kotlin/com/teamscale/TeamscalePluginTest.kt
```diff
@@ -135,7 +135,7 @@ class TeamscalePluginTest {
             .withProjectDir(temporaryFolder.root)
             .withPluginClasspath()
             .withArguments(runnerArgs)
-            .withGradleVersion("4.6")
+            .withGradleVersion("6.5")
```
DreierF, line 138: Why did you need to update the gradle version here?

4.6 is our minimal supported version and we run the tests against this version to ensure everything still works on this version.
When updating we should also adjust https://github.com/cqse/teamscale-jacoco-agent/blob/master/teamscale-gradle-plugin/src/main/kotlin/com/teamscale/TeamscalePlugin.kt#L50

DreierF (changes requested):

teamscale-gradle-plugin/src/test/kotlin/com/teamscale/TeamscalePluginTest.kt
karottenreibe: Ah I see. I was unable to run the tests on my machine since Gradle 4.6 seems to not run under Java 11, so I just updated the version to match what we use elsewhere. What's your opinion? Leave at 4.6 or bump to the first Gradle version that supports Java 11?

DreierF: Bump to Java 11 is fine for me. We are soon on 6.6.1 so 6.5 would be fine.

karottenreibe:
I added a commit that also includes the hostname in the default message

teamscale-gradle-plugin/src/test/kotlin/com/teamscale/TeamscalePluginTest.kt
DreierF: Could you please update the version in TeamscalePlugin (see link above)?

DreierF (approved):

PR #160, opened 2020-09-08
CR#24351 Add score and duration to API

DreierF:
Addresses issue [TS-24351](https://jira.cqse.eu/browse/TS-24351)

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ We use [semantic versioning](http://semver.org/):
 - [fix] option parsing errors were not logged in rare cases
 - [breaking change] changed default message to include partition and upload date
 - [breaking change] Bumped minimum supported Gradle version for the Gradle plugin to 6.5 so we can support Java 11
-- [feature] Made score and duration of test selected by TIA available via the tia-client
+- [feature] Made score and duration of test selected by TIA available via the teamscale-client
 
 # 18.0.0
 - [breaking change] removed options `coverage-via-http` and `teamscale-testwise-upload`. Use `tia-mode=http` or `tia-mode=teamscale-upload` instead.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ We use [semantic versioning](http://semver.org/):
 # Next Release
 
 - [fix] option parsing errors were not logged in rare cases
+- [feature] Made score and duration of test selected by TIA available via the tia-client
 
 # 18.0.0
 - [breaking change] removed options `coverage-via-http` and `teamscale-testwise-upload`. Use `tia-mode=http` or `tia-mode=teamscale-upload` instead.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ We use [semantic versioning](http://semver.org/):
 - [fix] option parsing errors were not logged in rare cases
 - [breaking change] changed default message to include partition and upload date
 - [breaking change] Bumped minimum supported Gradle version for the Gradle plugin to 6.5 so we can support Java 11
- 
+- [feature] Made score and duration of test selected by TIA available via the tia-client
```
karottenreibe, line 12: I think it's via the `teamscale-client` right?

teamscale-client/src/main/java/com/teamscale/client/PrioritizableTest.java
```diff
@@ -13,6 +15,18 @@
 	/** The reason the test has been selected. */
 	public String selectionReason;
 
+	/** Partition of the test. */
+	public String partition;
+
+	/** Duration in ms. May be null if not set. */
```
karottenreibe, line 21: I was wondering why it would not be set. I think this happens if the duration is not reported during the upload of the test-wise coverage, right? Should we mention this here explicitly so users that get `null` know what they have to do to get a sensible value here?

teamscale-client/src/main/java/com/teamscale/client/PrioritizableTest.java
```diff
@@ -13,6 +15,18 @@
...
+	public Long durationInMs;
+
+	/**
+	 * The score determined by the TIA algorithm.
```
karottenreibe, line 25: Please comment the semantics that we can guarantee. I think at least we should mention if higher = better score or lower = better score. Maybe there's more? Like the range 0-1 or 0-Infinity or something?

teamscale-client/src/main/java/com/teamscale/client/PrioritizableTestCluster.java
```diff
@@ -23,6 +22,12 @@
 	 */
 	public String clusterId;
 
+	/**
+	 * The score determined by the TIA algorithm for the cluster.
```
karottenreibe, line 26: see my comment on `score` in the other class

karottenreibe (changes requested):
minor comments

CHANGELOG.md
DreierF: Ups. Yes

teamscale-client/src/main/java/com/teamscale/client/PrioritizableTest.java
DreierF: Makes sense.

karottenreibe (approved):

PR #167, opened 2021-04-07
CR#26693 Allow multiple partitions

DreierF:
Addresses issue [TS-26693](https://cqse.atlassian.net/browse/TS-26693)

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ We use [semantic versioning](http://semver.org/):
 - PATCH version when you make backwards compatible bug fixes.
 
 # Next Release
+- [breaking change] _teamscale-client_: Impacted tests now allows to specify multiple partitions
+- [breaking change] _teamscale-client_: Impacted tests baseline can now be a string
+- [breaking change] _teamscale-client_: Impacted tests are queried with CHEAP_ADDITIONAL_COVERAGE_PER_TIME
 
 # 21.1.0
 
--- a/system-tests/default-excludes-test/src/test/java/com/teamscale/tia/client/TeamscaleMockServer.java
+++ b/system-tests/default-excludes-test/src/test/java/com/teamscale/tia/client/TeamscaleMockServer.java
@@ -30,7 +30,7 @@ public class TeamscaleMockServer {
 
 	public TeamscaleMockServer(int port) throws IOException {
 		port(port);
-		post("api/projects/:projectName/external-analysis/session/auto-create/report", this::handleReport);
+		post("api/v5.9.0/projects/:projectName/external-analysis/session/auto-create/report", this::handleReport);
 		exception(Exception.class, (Exception exception, Request request, Response response) -> {
 			response.status(SC_INTERNAL_SERVER_ERROR);
 			response.body("Exception: " + exception.getMessage());
--- a/teamscale-client/src/main/java/com/teamscale/client/ITeamscaleService.java
+++ b/teamscale-client/src/main/java/com/teamscale/client/ITeamscaleService.java
@@ -33,7 +33,7 @@ public interface ITeamscaleService {
 	 *     How to Upload External Analysis Results to Teamscale</a> for details.
 	 */
 	@Multipart
-	@POST("api/projects/{projectAliasOrId}/external-analysis/session/auto-create/report")
+	@POST("api/v5.9.0/projects/{projectAliasOrId}/external-analysis/session/auto-create/report")
 	Call<ResponseBody> uploadExternalReport(
 			@Path("projectAliasOrId") String projectAliasOrId,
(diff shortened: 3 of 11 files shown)


system-tests/default-excludes-test/src/test/java/com/teamscale/tia/client/TeamscaleMockServer.java
```diff
@@ -30,7 +30,7 @@
 
 	public TeamscaleMockServer(int port) throws IOException {
 		port(port);
-		post("api/projects/:projectName/external-analysis/session/auto-create/report", this::handleReport);
+		post("api/v5.9.0/projects/:projectName/external-analysis/session/auto-create/report", this::handleReport);
```
VitaNuova, line 33: Why are all these methods calling the APIs of specific Teamscale versions?

teamscale-client/src/main/java/com/teamscale/client/ITeamscaleService.java
```diff
@@ -81,33 +81,32 @@
...
 	/** Retrieve clustered impacted tests based on the given available tests and baseline timestamp. */
-	@PUT("api/projects/{projectName}/impacted-tests")
+	@PUT("api/v6.5.2/projects/{projectName}/impacted-tests?prioritization-strategy=CHEAP_ADDITIONAL_COVERAGE_PER_TIME")
 	Call<List<PrioritizableTestCluster>> getImpactedTests(
 			@Path("projectName") String projectName,
-			@Query("baseline") Long baseline,
+			@Query("baseline") String baseline,
 			@Query("end") CommitDescriptor end,
-			@Query("partitions") String partition,
+			@Query("partitions") List<String> partitions,
```
VitaNuova, line 89: Did you change the corresponding Teamscale service query option, or did it already support multiple partitions?

teamscale-client/src/main/java/com/teamscale/client/ITeamscaleService.java
```diff
@@ -81,33 +81,32 @@
 	);
 
 	/** Retrieve clustered impacted tests based on the given available tests and baseline timestamp. */
-	@PUT("api/projects/{projectName}/impacted-tests")
+	@PUT("api/v6.5.2/projects/{projectName}/impacted-tests?prioritization-strategy=CHEAP_ADDITIONAL_COVERAGE_PER_TIME")
```
VitaNuova, line 84: Is it fine that you're adding this prioritization strategy option to all calls to this endpoint, although it's only necessary for the current customer? Might we need another endpoint without this option, or might we need to parametrize this endpoint to be able to call with and without this strategy option?

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ We use [semantic versioning](http://semver.org/):
 - PATCH version when you make backwards compatible bug fixes.
 
 # Next Release
+- [breaking change] _teamscale-client_: Impacted tests now allows to specify multiple partitions
+- [breaking change] _teamscale-client_: Impacted tests baseline can now be a string
```
VitaNuova, line 9: What does this have to do with the customer request? And what's the benefit? Is that that you're now able to pass baseline names, like "Baseline Sprint 25"?

system-tests/default-excludes-test/src/test/java/com/teamscale/tia/client/TeamscaleMockServer.java
DreierF: Because otherwise we will break the agent once we introduce a breaking change to the API in the future. Instead when all external tools specify the version they expect (like the Java based IDEs already do) we can safely change stuff in the future without having to way years until all tools and instances are updated (as we previously had to)

DreierF: I mean isn't that the whole point of versioned APIs

teamscale-client/src/main/java/com/teamscale/client/ITeamscaleService.java
DreierF: It already did. We just didn't expose it here yet

teamscale-client/src/main/java/com/teamscale/client/ITeamscaleService.java
DreierF: This will become the new default soon. All other options are only for research purposes, but will not be used for production systems. So I thought it would be better to keep this as an implementation detail

CHANGELOG.md
DreierF: I missed to add that to the ticket description, but that was also requested. It's mostly to allow specifying 123456p1 baselines, but also to allow HEAD for example

PR #168, opened 2021-04-16
Fix spring boot

karottenreibe:
Addresses issue [TS-21994](https://jira.cqse.eu/browse/TS-21994)

Also fixes an issue with Spring boot that came up during the use of the new debug logs: Code locations start with jar:file instead of file: and thus couldn't be parsed to locate the jar files to search for git.properties.

- [x] Agent's README.md updated in case of user-visible changes - not needed
- [x] CHANGELOG.md updated
- [x] Present new features in [N&N](https://wiki.cqse.eu/pages/viewpage.action?pageId=689566) - not needed

### DIFF
--- a/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
@@ -1,7 +1,6 @@
 package com.teamscale.jacoco.agent.commit_resolution.git_properties;
 
 import com.teamscale.client.StringUtils;
-import com.teamscale.jacoco.agent.options.AgentOptions;
 import com.teamscale.jacoco.agent.options.ProjectRevision;
 import com.teamscale.report.util.BashFileSkippingInputStream;
 import org.conqat.lib.commons.collections.Pair;
@@ -9,6 +8,8 @@ import org.conqat.lib.commons.collections.Pair;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.nio.file.Paths;
 import java.util.Properties;
 import java.util.jar.JarEntry;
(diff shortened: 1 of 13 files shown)


agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -42,24 +43,47 @@ public static String getRevisionFromGitProperties(File jarFile) throws IOExcepti
...
+	 */
+	public static File extractGitPropertiesSearchRoot(URL jarOrClassFolderUrl) throws URISyntaxException {
+		String protocol = jarOrClassFolderUrl.getProtocol().toLowerCase();
+		if (protocol.equals("file") && org.conqat.lib.commons.string.StringUtils.endsWithOneOf(
+				jarOrClassFolderUrl.getPath().toLowerCase(), ".jar", ".war", ".ear", ".aar")) {
+			return new File(jarOrClassFolderUrl.toURI());
+		} else if (protocol.equals("jar")) {
+			// used e.g. by Spring Boot. Example: jar:file:/home/k/demo.jar!/BOOT-INF/classes!/
+			String urlString = jarOrClassFolderUrl.toString();
+			String path = urlString.substring(9, urlString.indexOf("!/"));
```
VitaNuova, line 61: Magic number. Possibly better `urlString.indexOf(":")` (in case my eye counting is correct ;-) ?

agent/src/test/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtilsTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.io.File;
+import java.net.URL;
+
+class GitPropertiesLocatorUtilsTest {
+
+	@Test
+	public void parseSpringBootCodeLocations() throws Exception {
+		Assertions.assertThat(GitPropertiesLocatorUtils
+				.extractGitPropertiesSearchRoot(new URL("jar:file:/home/k/demo.jar!/BOOT-INF/classes!/")))
```
VitaNuova, line 14: Extract `/home/k/demo.jar` to a constant? It's used four times in this class AFAICS.

report-generator/src/main/java/com/teamscale/report/jacoco/TeamscaleCoverageBuilder.java
```diff
@@ -17,9 +16,9 @@
 import java.util.Collections;
 
 /**
- * Modified {@link CoverageBuilder} can ignore non-identical duplicate classes or classes without coverage.
- * In addition, Coverage returned via {@link #getBundle(String)} will only return source file coverage,
- * because Teamscale does not need class coverage anyway. This reduces XML size by approximately half.
+ * Modified {@link CoverageBuilder} can ignore non-identical duplicate classes or classes without coverage. In addition,
+ * Coverage returned via {@link #getBundle(String)} will only return source file coverage, because Teamscale does not
```
VitaNuova, line 20:
```suggestion
 * coverage returned via {@link #getBundle(String)} will only return source file coverage because Teamscale does not
```

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
karottenreibe: Yeah, I was lazy. I solved it with a regex, which I think is the best solution. That documents the full format of the URL we're parsing

karottenreibe: Plus it works on Windows where I think the path doesn't start with a slash but with "C:" etc.

agent/src/test/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtilsTest.java
karottenreibe: I'd like to keep the tests independent. Changing one of them should not affect the others. And I find it much easier to read the test when the string is not chopped up into `"jar:file:" + TEST_JAR_PATH + "!/BOOT-INF/classes"`. I also don't think that renaming that file consistently across all tests is really a maintenance use case we should be optimizing for over the plain readability of the test.

VitaNuova (approved):

PR #173, opened 2021-05-03
CR#26983 Delay testwise coverage generation to testrun/end and theref…

DreierF:
Delay testwise coverage generation to testrun/end and therefore make it reasonable to rescan the classes dir on report generation

Addresses issue [TS-26983](https://cqse.atlassian.net/browse/TS-26983)

- [ ] CHANGELOG.md updated

### DIFF
--- a/agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
@@ -38,6 +38,7 @@ import org.jacoco.core.runtime.WildcardMatcher;
 import org.slf4j.Logger;
 
 import java.io.File;
+import java.io.IOException;
 import java.lang.instrument.Instrumentation;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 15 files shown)


agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
```diff
@@ -426,6 +427,12 @@ public Path getOutputDirectory() {
 		return outputDirectory;
 	}
 
+	/** Creates a new temp file with the given prefix, extension and current timestamp and ensures that the parent folder actually exists. */
```
Raphael-N, line 430: It sounds like this would be better placed in FileSystemUtils or something similar. What do you think?

DreierF: In general I agree, but there is the outputDirectory and the way the file name is built up that I would both consider specific to the agent and therefore not suitable for a general purpose method in utils.

And the ensureDirectoryExists already is implemented in the general purpose part.

Would you instead prefer having a `FileSystemUtils.createTempFileWithTimestamp(Path parentDir, String prefix, String extension)`?

But I found even two more occurrences where replacing with `createTempFile` is possible.

Raphael-N: Ok, it's fine with me then.

PR #175, opened 2021-05-17
CR#26966 Implement vfs support

VitaNuova:
Addresses issue [TS-XXXXX](https://cqse.atlassian.net/browse/TS-26966)

- [x] CHANGELOG.md updated

### DIFF
--- a/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitMultiProjectPropertiesLocator.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitMultiProjectPropertiesLocator.java
@@ -42,14 +42,14 @@ public class GitMultiProjectPropertiesLocator implements IGitPropertiesLocator {
 	 * Asynchronously searches the given jar file for a git.properties file.
 	 */
 	@Override
-	public void searchJarFileForGitPropertiesAsync(Pair<File, Boolean> file) {
-		executor.execute(() -> searchJarFile(file));
+	public void searchFileForGitPropertiesAsync(File file, boolean isJarFile) {
+		executor.execute(() -> searchFile(file, isJarFile));
 	}
 
-	private void searchJarFile(Pair<File, Boolean> file) {
-		logger.debug("Searching jar file {} for multiple git.properties", file.toString());
+	private void searchFile(File file, boolean isJarFile) {
+		logger.debug("Searching file {} for multiple git.properties", file.toString());
 		try {
-			ProjectRevision data = GitPropertiesLocatorUtils.getProjectRevisionFromGitProperties(file.getFirst(), file.getSecond());
+			ProjectRevision data = GitPropertiesLocatorUtils.getProjectRevisionFromGitProperties(file, isJarFile);
 			if (data == null) {
 				logger.debug("No git.properties file found in {}", file.toString());
 				return;
--- a/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitMultiProjectPropertiesLocator.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitMultiProjectPropertiesLocator.java
@@ -4,7 +4,6 @@ import com.teamscale.jacoco.agent.options.ProjectRevision;
 import com.teamscale.jacoco.agent.upload.teamscale.DelayedTeamscaleMultiProjectUploader;
 import com.teamscale.jacoco.agent.util.DaemonThreadFactory;
 import com.teamscale.jacoco.agent.util.LoggingUtils;
-import org.conqat.lib.commons.collections.Pair;
 import org.slf4j.Logger;
 
 import java.io.File;
--- a/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitMultiProjectPropertiesLocator.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitMultiProjectPropertiesLocator.java
@@ -4,6 +4,7 @@ import com.teamscale.jacoco.agent.options.ProjectRevision;
 import com.teamscale.jacoco.agent.upload.teamscale.DelayedTeamscaleMultiProjectUploader;
 import com.teamscale.jacoco.agent.util.DaemonThreadFactory;
 import com.teamscale.jacoco.agent.util.LoggingUtils;
+import org.conqat.lib.commons.collections.Pair;
 import org.slf4j.Logger;
 
 import java.io.File;
@@ -41,16 +42,16 @@ public class GitMultiProjectPropertiesLocator implements IGitPropertiesLocator {
 	 * Asynchronously searches the given jar file for a git.properties file.
 	 */
 	@Override
-	public void searchJarFileForGitPropertiesAsync(File jarFile) {
-		executor.execute(() -> searchJarFile(jarFile));
+	public void searchJarFileForGitPropertiesAsync(Pair<File, Boolean> file) {
+		executor.execute(() -> searchJarFile(file));
 	}
 
-	private void searchJarFile(File jarFile) {
-		logger.debug("Searching jar file {} for multiple git.properties", jarFile.toString());
+	private void searchJarFile(Pair<File, Boolean> file) {
+		logger.debug("Searching jar file {} for multiple git.properties", file.toString());
 		try {
-			ProjectRevision data = GitPropertiesLocatorUtils.getProjectRevisionFromGitProperties(jarFile);
+			ProjectRevision data = GitPropertiesLocatorUtils.getProjectRevisionFromGitProperties(file.getFirst(), file.getSecond());
 			if (data == null) {
-				logger.debug("No git.properties file found in {}", jarFile.toString());
+				logger.debug("No git.properties file found in {}", file.toString());
 				return;
 			}
 			// this code only runs when 'teamscale-project' is not given via the agent properties,
@@ -62,14 +63,14 @@ public class GitMultiProjectPropertiesLocator implements IGitPropertiesLocator {
 						"Found inconsistent git.properties file: the git.properties file in {} either does not specify the" +
 								" Teamscale project (teamscale.project) property, or does not specify the commit SHA (git.commit.id)." +
 								" Please note that both of these properties are required in order to allow multi-project upload to Teamscale.",
-						jarFile);
+						file);
 				return;
 			}
 			uploader.setTeamscaleProjectForRevision(data);
-			logger.debug("Found git.properties file in {} and found Teamscale project {} and revision {}", jarFile.toString(),
+			logger.debug("Found git.properties file in {} and found Teamscale project {} and revision {}", file.toString(),
 					data.getProject(), data.getRevision());
 		} catch (IOException | InvalidGitPropertiesException e) {
-			logger.error("Error during asynchronous search for git.properties in {}", jarFile.toString(), e);
+			logger.error("Error during asynchronous search for git.properties in {}", file.toString(), e);
 		}
 	}
 
--- a/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocator.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocator.java
@@ -3,6 +3,7 @@ package com.teamscale.jacoco.agent.commit_resolution.git_properties;
 import com.teamscale.jacoco.agent.upload.delay.DelayedUploader;
 import com.teamscale.jacoco.agent.util.DaemonThreadFactory;
 import com.teamscale.jacoco.agent.util.LoggingUtils;
+import org.conqat.lib.commons.collections.Pair;
 import org.slf4j.Logger;
 
 import java.io.File;
@@ -51,16 +52,16 @@ public class GitPropertiesLocator<T> implements IGitPropertiesLocator {
 	 * Asynchronously searches the given jar file for a git.properties file.
 	 */
 	@Override
-	public void searchJarFileForGitPropertiesAsync(File jarFile) {
-		executor.execute(() -> searchJarFile(jarFile));
+	public void searchJarFileForGitPropertiesAsync(Pair<File, Boolean> file) {
+		executor.execute(() -> searchJarFile(file));
 	}
 
-	private void searchJarFile(File jarFile) {
-		logger.debug("Searching jar file {} for a single git.properties", jarFile);
+	private void searchJarFile(Pair<File, Boolean> file) {
+		logger.debug("Searching jar file {} for a single git.properties", file);
 		try {
-			T data = dataExtractor.extractData(jarFile);
+			T data = dataExtractor.extractData(file.getFirst(), file.getSecond());
 			if (data == null) {
-				logger.debug("No git.properties file found in {}", jarFile.toString());
+				logger.debug("No git.properties file found in {}", file.toString());
 				return;
 			}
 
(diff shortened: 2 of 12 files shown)


agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -55,22 +59,50 @@ public static String getRevisionFromGitProperties(File jarFile) throws IOExcepti
...
+		// used by JBoss EAP and Wildfly. Example: vfs:/content/helloworld.war/WEB-INF/classes
+		Object virtualFile = jarOrClassFolderUrl.openConnection().getContent();
+		Class<?> virtualFileClass = virtualFile.getClass();
+		// obtain the physical location of the class file. It is created on demand in <jboss-installation-dir>/standalone/tmp/vfs
+		Method getPhysicalFileMethod = virtualFileClass.getMethod("getPhysicalFile");
+		File file = (File) getPhysicalFileMethod.invoke(virtualFile);
+		if (file.getParentFile() == null || file.getParentFile().getParentFile() == null) {
+			return null;
+		}
+		// the physical location of the class is in, e.g., WEB-INF/classes folder.
```
karottenreibe, line 99: I don't think this will work in all cases. the URLs I've seen from our customer did not look like this but were at a deeper nesting level. I think we instead have to manipulate the URL so it points to `helloworld.war` and then resolve that. I.e. extract the first war/jar/ear file from the url, throw away the rest of the url path and resolve the jar file to disk

karottenreibe (commented):
one suggestion, since I read your comment about brittleness in Jira @VitaNuova

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/IGitPropertiesLocator.java
```diff
@@ -1,10 +1,12 @@
...
+import org.conqat.lib.commons.collections.Pair;
+
 import java.io.File;
 
 /** Interface for the locator classes that search JAR files for git.properties files containing certain properties. */
 public interface IGitPropertiesLocator {
 
 	/** Searches the JAR file for the git.properties file containing certain properties. */
-	void searchJarFileForGitPropertiesAsync(File jarFile);
+	void searchJarFileForGitPropertiesAsync(Pair<File, Boolean> jarFile);
```
DreierF, line 11: The doc comment should mention what the pair represents. For a plain file this was quite obvious, but the boolean flag needs more context.

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitMultiProjectPropertiesLocator.java
```diff
@@ -41,16 +42,16 @@ public GitMultiProjectPropertiesLocator(DelayedTeamscaleMultiProjectUploader upl
...
-	private void searchJarFile(File jarFile) {
-		logger.debug("Searching jar file {} for multiple git.properties", jarFile.toString());
+	private void searchJarFile(Pair<File, Boolean> file) {
+		logger.debug("Searching jar file {} for multiple git.properties", file.toString());
 		try {
-			ProjectRevision data = GitPropertiesLocatorUtils.getProjectRevisionFromGitProperties(jarFile);
+			ProjectRevision data = GitPropertiesLocatorUtils.getProjectRevisionFromGitProperties(file.getFirst(), file.getSecond());
 			if (data == null) {
-				logger.debug("No git.properties file found in {}", jarFile.toString());
+				logger.debug("No git.properties file found in {}", file.toString());
```
DreierF, line 54: Shouldn't that become

```suggestion
				logger.debug("No git.properties file found in {}", file.getFirst().toString());
```

I would also suggest to rename file to something that better describes what the pair represents

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitMultiProjectPropertiesLocator.java
```diff
@@ -41,16 +42,16 @@ public GitMultiProjectPropertiesLocator(DelayedTeamscaleMultiProjectUploader upl
 	 * Asynchronously searches the given jar file for a git.properties file.
 	 */
 	@Override
-	public void searchJarFileForGitPropertiesAsync(File jarFile) {
-		executor.execute(() -> searchJarFile(jarFile));
+	public void searchJarFileForGitPropertiesAsync(Pair<File, Boolean> file) {
```
DreierF, line 45: Below there are quite some logging calls that should now use file.getFirst() instead. WDYT of making this two arguments instead of one pair so that you can give them descriptive names?

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocator.java
```diff
@@ -51,16 +52,16 @@ public GitPropertiesLocator(DelayedUploader<T> uploader, DataExtractor<T> dataEx
...
 	@Override
-	public void searchJarFileForGitPropertiesAsync(File jarFile) {
-		executor.execute(() -> searchJarFile(jarFile));
+	public void searchJarFileForGitPropertiesAsync(Pair<File, Boolean> file) {
+		executor.execute(() -> searchJarFile(file));
 	}
 
-	private void searchJarFile(File jarFile) {
-		logger.debug("Searching jar file {} for a single git.properties", jarFile);
+	private void searchJarFile(Pair<File, Boolean> file) {
```
DreierF, line 59: Maybe also split up the pair into two parameters here

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -97,19 +129,47 @@ public static ProjectRevision getProjectRevisionFromGitProperties(
...
+			return findGitPropertiesInJarFile(jarStream);
 		} catch (IOException e) {
-			throw new IOException("Reading jar " + jarFile.getAbsolutePath() + " for obtaining commit " +
+			throw new IOException("Reading jar " + file.getAbsolutePath() + " for obtaining commit " +
 					"descriptor from git.properties failed", e);
 		}
 	}
 
+	private static Pair<String, Properties> findGitPropertiesInDirectoryFile(File file) throws IOException {
+		File[] directoryFiles = file.listFiles();
```
DreierF, line 151: That will only list direct children, right? I guess we need a recursive solution here. 

`FileSystemUtils.listFilesRecursively` might be of help here

DreierF (changes requested):

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
VitaNuova: Ok makes sense, thanks.

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
VitaNuova: Yes, I have actually assumed the git.properties file should always be at the top. If this is not the case, then we need the recursive solution.

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -85,22 +87,48 @@ public static String getRevisionFromGitProperties(
...
+	}
+
+	private static String extractArtefactUrl(URL jarOrClassFolderUrl) {
+		String url = jarOrClassFolderUrl.getPath().toLowerCase();
+		String[] pathSegments = url.split("/");
+		StringBuilder artefactUrlBuilder = new StringBuilder("vfs:");
+		int segmentIdx = 0;
+		String segment = pathSegments[segmentIdx];
+		while (segmentIdx < pathSegments.length && !org.conqat.lib.commons.string.StringUtils.endsWithOneOf(
+				segment, ".jar", ".war", ".ear", ".aar")) {
```
DreierF, line 119: I would suggest to move the second condition into the loop as an if statement that will break if the condition is fulfilled. Then you also do not need to append `pathSegments[segmentIdx]` afterwards. I also think you are currently missing an additional slash in that case, right?

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -85,22 +87,48 @@ public static String getRevisionFromGitProperties(
...
+		Object virtualFile = new URL(artefactUrl).openConnection().getContent();
 		Class<?> virtualFileClass = virtualFile.getClass();
 		// obtain the physical location of the class file. It is created on demand in <jboss-installation-dir>/standalone/tmp/vfs
 		Method getPhysicalFileMethod = virtualFileClass.getMethod("getPhysicalFile");
 		File file = (File) getPhysicalFileMethod.invoke(virtualFile);
-		if (file.getParentFile() == null || file.getParentFile().getParentFile() == null) {
+		return Pair.createPair(file, false);
+	}
+
+	private static String extractArtefactUrl(URL jarOrClassFolderUrl) {
```
DreierF, line 112: IMO it would be good to have a unit test for that

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -85,22 +87,48 @@ public static String getRevisionFromGitProperties(
...
+				segment, ".jar", ".war", ".ear", ".aar")) {
+			artefactUrlBuilder.append(segment);
+			artefactUrlBuilder.append("/");
+			if (segmentIdx + 1 < pathSegments.length) {
+				segment = pathSegments[segmentIdx + 1];
+			}
+			segmentIdx += 1;
+		}
+		if (segmentIdx == pathSegments.length) {
 			return null;
```
DreierF, line 128: Do we really want to return null here? I would have expected the full path. Not sure if that is possible in JBoss, but I could image that you can besides jar's war's etc. also add folders of class files which would presumably end up as vfs:/content/path/to/my/folder

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -147,25 +175,23 @@ public static ProjectRevision getProjectRevisionFromGitProperties(
...
-	private static Pair<String, Properties> findGitPropertiesInDirectoryFile(File file) throws IOException {
-		File[] directoryFiles = file.listFiles();
-		if (directoryFiles == null) {
-			return null;
+	private static Pair<String, Properties> findGitPropertiesInDirectoryFile(File directoryFile) throws IOException {
+		List<File> directoryFiles = FileSystemUtils.listFilesRecursively(directoryFile,
+				file -> file.getName().toLowerCase().equals(GIT_PROPERTIES_FILE_NAME));
+		if (directoryFiles.size() != 1) {
+			throw new IllegalArgumentException(
+					"There must be a single git.properties file in the directory " + directoryFile.toString());
```
DreierF, line 183: Would help in debugging if we would include the paths under which we found git.properties as the files could get quite big.

Must there be a git.properties in every scanned folder path? Or could there be multiple with only a single git.properties?

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/IGitPropertiesLocator.java
```diff
@@ -4,9 +4,12 @@
...
 import java.io.File;
 
-/** Interface for the locator classes that search JAR files for git.properties files containing certain properties. */
+/** Interface for the locator classes that search files (e.g., a JAR) for git.properties files containing certain properties. */
 public interface IGitPropertiesLocator {
 
-	/** Searches the JAR file for the git.properties file containing certain properties. */
-	void searchJarFileForGitPropertiesAsync(Pair<File, Boolean> jarFile);
+	/**
+	 * Searches the file for the git.properties file containing certain properties. The boolean frag indicates whether the
```
DreierF, line 11:
```suggestion
	 * Searches the file for the git.properties file containing certain properties. The boolean flag indicates whether the
```

DreierF (changes requested):
Nice improvement!

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
VitaNuova: Yes, I am deliberately missing it. Why would we need a slash at the end?

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
VitaNuova: What do you mean by "add folders of class files"? You mean deploy class files to the application server directly, without a jar/war artefact? I don't know if something like this is possible. However, if I return the full path where the class is residing, then further git.properties search will only be done relative to this path, is that fine?

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
VitaNuova: Ah, you're right. I forgot it might simply be a jar/war with no git.properties inside. Then we just scan the next.

However, I would still expect a single git.properties file to be present in a single jar/war, if any. Otherwise, we wouldn't know which values to adopt, which is a problem.

VitaNuova: I don't quite get the comment regarding debugging. Do you really want to me log the exact locations in case we find multiple git.properties files in a single jar? This is both 
 vary unlikely and can get quite lengthy.

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
VitaNuova: ```java.net.MalformedURLException: unknown protocol: vfs```
I cannot artificially construct a VFS URL. It's probably possible using the `org.jboss.vfs` library, but since we're doing everything with reflection now, there is no dependency anymore. I guess we don't want an extra dependency for the sake of this test case.

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
DreierF: Yes, that is what I meant. In case you really run into such a problem I would rather have too much information then not knowing what is going on. In case you really end up with two git.properties you most likely did not expect this to happen and it makes it easier to spot the error if you see the two paths side by side that the agent found.

VitaNuova: Done.

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
DreierF: Not at the end. Between the last Segment and the Segment before that

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
DreierF: The alternative would be to skip looking in that location entirely, right?

DreierF (changes requested):
Sorry one last point: we are still missing a chance log entry

PR #180, opened 2021-07-29
Fix Wildfly docs regarding domain mode

karottenreibe:


agent/README.md
```diff
@@ -348,12 +348,28 @@ This ensures that the performance of your application does not degrade.
...
 
-Register the agent in the `JAVA_OPTS` environment variable in the `standalone.conf` or `domain.conf`
-file inside the Wildfly installation directory - depending on which "mode" is used; probably standalone.
-
 Please set the agent's `includes` parameter so that the Wildfly code is not being profiled.
 This ensures that the performance of your application does not degrade.
 
+### Using standalone mode
+
+Register the agent in the `JAVA_OPTS` environment variable in the `standalone.conf`
```
stahlbauer, line 356: Besides `domain.xml` there is a dual `standalone.xml`. Remove the JAVA_OPTS-Approach? Keep it for other/older versions?

agent/README.md
```diff
@@ -348,12 +348,28 @@ This ensures that the performance of your application does not degrade.
...
+file inside the Wildfly installation directory.
+
+### Using domain mode
+
+Register the agent in `domain.xml` under
```
stahlbauer, line 361: Make the same for `standalone.xml`?

stahlbauer (changes requested):

agent/README.md
karottenreibe: I would not change the standalone stuff unless we are sure that it needs changing. The last Wildfly I instrumented was in standalone and definitely required changing the conf file

agent/README.md
karottenreibe: See other comment :-)

PR #207, opened 2022-04-26
TS-23914 Detect and warn about other agents.

Rixtho:

- [ ] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/AgentBase.java
```diff
@@ -119,6 +121,14 @@ protected void initServerEndpoints(Service spark) {
 	public static void premain(String options, Instrumentation instrumentation) throws Exception {
 		AgentOptions agentOptions;
 		DelayedLogger delayedLogger = new DelayedLogger();
+
+		List<String> javagents = CollectionUtils.filter(ManagementFactory.getRuntimeMXBean().getInputArguments(),
+				s -> s.contains("-javaagent"));
+		if (javagents.size() > 1 || !javagents.get(0).contains("teamscale-jacoco-agent.jar")) {
```
andipunz, line 127: Maybe we should split this logging into two cases:
1. Multiple Agents -> Warning with message multiple agents
2. If multiple agents AND Jacoco not the first one hint that for best results the jacoco agent should be the first one

PR #208, opened 2022-04-28
TS-21992 Make Debug Logging Easier

Rixtho:
Addresses issue [TS-21992](https://cqse.atlassian.net/browse/TS-21992)

- [ ] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
```diff
@@ -79,6 +79,12 @@ public class AgentOptions {
 	 */
 	/* package */ String originalOptionsString;
 
+	/** Whether debug logging is active or not. */
+	/* package */ boolean debugLogging = false;
+
+	/** Explicitly defined log file. */
+	/* package */ String loggingFilePath = "";
```
andipunz, line 86: Is this just working for debug logging or for standard logging as well?

andipunz, line 86:
```suggestion
	/* package */ String debugLoggingFilePath = "";
```

agent/src/main/java/com/teamscale/jacoco/agent/util/DebugLogDirectoryPropertyDefiner.java
```diff
@@ -0,0 +1,17 @@
...
+
+	/* package */ static Path filePath = null;
+
+	@Override
+	public String getPropertyValue() {
+		if (filePath == null) {
+			return super.getPropertyValue();
+		}
+		return filePath.resolve("logs").toAbsolutePath().toString();
+	}
```
andipunz, line 16: Not sure if this is the right place to do it, but we should handle the case when its either a wrong path or the agent is not allowed to log to the path

andipunz (changes requested):

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
Rixtho: Only for debug logging

agent/src/main/java/com/teamscale/jacoco/agent/util/DebugLogDirectoryPropertyDefiner.java
Rixtho: I added a check whether the file path is valid, but how can I check whether the file path can be written to?

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
andipunz: Then I would rename to debugLoggingFilePath to make it clear

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
```diff
@@ -188,6 +194,16 @@ public AgentOptions() {
...
 
+	/** @see #debugLogging */
+	public boolean isDebugLogging() {
+		return debugLogging;
+	}
+
+	/** @see #loggingFilePath */
+	public String getLoggingFilePath() {
+		return loggingFilePath;
+	}
```
andipunz, line 205: See my comment above

agent/src/main/java/com/teamscale/jacoco/agent/util/DebugLogDirectoryPropertyDefiner.java
andipunz: Not sure about that rn. Maybe thats to much of an overhead. Can you check whats happening when you define a not existing or not writeable path? If thats handled gracefully we can skip the validation here

Rixtho: In case of an invalid file path, we print the exception to the console and then continue running normally. But logging is then only done to the console.
Another thing we can do is pass null as the logging path. The logger will then create a new directory named `debugLogDirectory_IS_UNDEFINED`. Which one do you prefer?

andipunz: The first option is ok

agent/src/main/java/com/teamscale/jacoco/agent/AgentBase.java
```diff
@@ -135,8 +139,18 @@ public static void premain(String options, Instrumentation instrumentation) thro
...
+			Path logDirectory = agentOptions.getDebugLogDirectory().resolve("logs");
+			if (FileSystemUtils.isValidPath(logDirectory.toString()) && Files.isWritable(logDirectory)) {
+				delayedLogger.info("Logging to " + logDirectory);
+			} else {
+				delayedLogger.warn("Could not create " + logDirectory + ". Logging to console only.");
+			}
+		} else {
+			loggingResources = LoggingUtils.initializeLogging(agentOptions.getLoggingConfig());
+			delayedLogger.info("Logging to " + new LogDirectoryPropertyDefiner().getPropertyValue());
+		}
```
andipunz, line 153: Can you extract this to a own method in order to resolve the Teamscale finding

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptionsParser.java
```diff
@@ -92,6 +92,13 @@ private void handleOption(AgentOptions options, String optionPart) throws AgentO
 		Pair<String, String> keyAndValue = parseOption(optionPart);
 		String key = keyAndValue.getFirst();
 		String value = keyAndValue.getSecond();
+		if (key.startsWith("debug")) {
+			options.debugLogging = true;
+			if (!value.isEmpty() && !value.equalsIgnoreCase("true")) {
+				options.debugLogDirectory = Paths.get(value);
+			}
+			return;
+		}
```
andipunz, line 101: Can you extract this to a own method to remove the Teamscale finding and to make it look cleaner here, especially with the nesting if

PR #211, opened 2022-05-04
TS-30166 standard artifact storage schema

andipunz:
Addresses issue [TS-30166](https://cqse.atlassian.net/browse/TS-30166)

- [x] CHANGELOG.md updated


agent/README.md
```diff
@@ -36,164 +35,178 @@ The following options are available:
...
-   If not specified, Git commit information is extracted from the first found `git.properties` file.
-   See `teamscale-git-properties-jar` for details. 
+  a user with write access (
+  c.f. [Artifactory Documentation](https://www.jfrog.com/confluence/display/JFROG/User+Profile#UserProfile-APIKey))
+- `artifactory-zip-path` (optional): The path within the stored ZIP file where the reports are stored. Default is to
+  store at root level. This can be used to encode e.g. a partition name that is parsed later on via Teamscale
+  Artifactory connector options.
+- `artifactory-path-suffix` (optional): The path within the storage location between
+  the `<root-path>/uploads/<branch name>/<commit timestamp>[-<commit hash>]?/<upload partition>/<artifact type>` and the
+  uploaded artifact.
```
DreierF, line 188: It is unclear to me what kind of value I should pass in as a user here. Is it `uploads`? Is it the whole pattern that is shown here? Is it something that would come even after that presented pattern?

On a related topic: We should document somewhere (CHANGELOG) maybe how the upgrade path for existing users looks like. Do we recommend to adjust the configuration in Teamscale and use the new default format? How would the user need to configure the agent to keep the old behavior (in the valid case that the customer does not want to loose all old uploads)...

DreierF (changes requested):

agent/README.md
andipunz: Regarding the second comment: Currently it's not possible to edit the default. There's a feature ticket in the backlog but it's not implemented yet.

DreierF: @karottenreibe What is the strategy from your point of view that we should push towards? I would just want to avoid that this introduces such a breaking change that existing users cannot use any newer version of the agent without loosing all previous data.

agent/README.md
```diff
@@ -36,164 +35,178 @@ The following options are available:
...
+      plain text.
     - `[GET] /commit` Returns the current commit used for uploading to Teamscale.
-    - `[PUT] /commit` Sets the commit to use for uploading to Teamscale. The commit must be in the request body in plain thext in the format: branch:timestmap
-- `artifactory-url`: the HTTP(S) url of the artifactory server to upload the reports to.
-   The URL may include a subpath on the artifactory server, e.g. `https://artifactory.acme.com/my-repo/my/subpath`.
+    - `[PUT] /commit` Sets the commit to use for uploading to Teamscale. The commit must be in the request body in plain
+      thext in the format: branch:timestmap
+- `artifactory-url`: the HTTP(S) url of the artifactory server to upload the reports to. The URL may include a subpath
+  on the artifactory server, e.g. `https://artifactory.acme.com/my-repo/my/subpath`.
+- `artifactory-partition`: the partition name that is parsed later on via Teamscale Artifactory connector options.
```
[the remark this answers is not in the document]
DreierF: I think I misread this sentence the first time. Sorry 🙈  I guess parsed was the correct term. Without the context it seemed like `pass on sth.`

agent/README.md
DreierF: @andipunz We definitely need to state somewhere what this "default path" actually is and how to configure it in Teamscale manually even though it will be the default in the next Teamscale versions. Without this information it is impossible (without reading the code) to find out how to fit those pieces together.

andipunz: I implemented a legacy-upload option in order to keep the new behavior. I've just checked the issues regarding the support in TS itself. They are still in dev. Should we provide somewhere in the docs (here? TS-docs) how the extraction pattern needs to be modified to use this new version?

agent/src/main/java/com/teamscale/jacoco/agent/upload/artifactory/ArtifactoryConfig.java
```diff
@@ -118,6 +127,13 @@ public static boolean handleArtifactoryOptions(ArtifactoryConfig options,
 			case ARTIFACTORY_PASSWORD_OPTION:
 				options.password = value;
 				return true;
+			case ARTIFACTORY_LEGACY_PATH_OPTION:
+				if (value.isEmpty()) {
```
DreierF, line 131: I would suggest to get rid of this special handling to be consistent with all other boolean options in the agent i.e. `dump-on-exit`, `ignore-uncovered-classes`, `validate-ssl`, ...
IMO it's less confusing to give the user one clear option instead of two and also it would feel strange to omit the value in the properties file.

agent/README.md
DreierF: Definitely yes. (See my other comment for more details)

DreierF (changes requested):

agent/README.md
```diff
@@ -184,12 +195,42 @@ directories, you can get the commit info via
...
+
+### The new standard upload schema
+```
+/ arbitraryly / many / directories / before / the / upload / root
+    / 'uploads'
+    / <branch name> 
+    / <commit timestamp (Java long)>[-<commit hash>]?
+    / <upload partition> 
+    / <artifact type> 
+    / none / or / some / internediate / directories /
```
DreierF, line 207:
```suggestion
    / none / or / some / intermediate / directories /
```

agent/README.md
```diff
@@ -184,12 +195,42 @@ directories, you can get the commit info via
...
+### The legacy standard upload schema
+```
+/ arbitraryly / many / directories / before / the / upload / root
+    / <branch name> 
+    / <commit timestamp (Java long)>[-<commit hash>]?
+    / <Zip archive file>
+```
+
+### Migration from the legacy to the new standard upload schema
+- Add a second artifactory connector to the Teamscale project with the new pattern next to the old one.
```
DreierF, line 229:
```suggestion
- Add a second artifactory connector to the Teamscale project with the new patterns next to the already existing Artifactory connector.
```

agent/README.md
```diff
@@ -75,128 +75,170 @@ The log file is written to the agent's directory in the subdirectory `logs` by d
...
+  class) and a Teamscale project alias or ID separated by a colon. The marker class must be guaranteed to be executed
+  when the application is running and is unique amongst the other deployed applications.
+  E.g. `com.company.app1.Main:app1alias;com.company.app2.Starter:ts-app2-id`. The coverage is uploaded to master at
+  the timestamp of the last modification date of the given marker class.
+
+### Options for the Artifactory Upload
 - `artifactory-url`: the HTTP(S) url of the artifactory server to upload the reports to.
-   The URL may include a subpath on the artifactory server, e.g. `https://artifactory.acme.com/my-repo/my/subpath`.
+  The URL may include a subpath on the artifactory server, e.g. `https://artifactory.acme.com/my-repo/my/subpath`.
+- `artifactory-legacy-path`: option to use the upload path as defined pre v24.0.0. Use `-artifactory-legacy-path=true` to keep the old behavior. Defaults to false. For more details see [new upload schema](#the-new-standard-upload-schema), [legacy upload schema](#the-legacy-standard-upload-schema) and [migration](#migration-from-the-legacy-to-the-new-standard-upload-schema).
```
andipunz, line 182:
```suggestion
- `artifactory-legacy-path`: option to use the upload path as defined pre v24.0.0. Use `artifactory-legacy-path=true` to keep the old behavior. Defaults to false. For more details see [new upload schema](#the-new-standard-upload-schema), [legacy upload schema](#the-legacy-standard-upload-schema) and [migration](#migration-from-the-legacy-to-the-new-standard-upload-schema).
```

PR #218, opened 2022-07-20
Cr/30151 finding git properties recursively

AnonymFx:
Addresses issue [TS-30151](https://cqse.atlassian.net/browse/TS-30151)

- [x] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocator.java
```diff
@@ -58,14 +64,21 @@ public void searchFileForGitPropertiesAsync(File file, boolean isJarFile) {
...
+						". Uploading to multiple projects is not possible with the option teamscale-project.");
+
 				return;
 			}
+			T dataEntry = data.get(0);
 
 			if (foundData != null) {
-				if (!foundData.equals(data)) {
+				if (!foundData.equals(dataEntry)) {
 					logger.error(
```
Raphael-N, line 82: This seems not fully consistent with your change from above. You say that `if (data.size() > 1)` we don't look at either of the git.properties files and return instead. Here, we log an error and keep using the first one we found. Also, do you know, in what scenario we can find another one here?

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -37,12 +38,9 @@ public class GitPropertiesLocatorUtils {
 	/** Matches the path to the jar file in a jar:file: URL in regex group 1. */
 	private static final Pattern JAR_URL_REGEX = Pattern.compile("jar:file:(.*?)!/.*", Pattern.CASE_INSENSITIVE);
 
-	private static final Pattern NESTED_JAR_REGEX = Pattern.compile("[j|w]ar:file:(.*?)\\*(.*)",
+	private static final Pattern NESTED_JAR_REGEX = Pattern.compile("[jwea]ar:file:(.*?)\\*(.*)",
```
Raphael-N, line 41: Maybe rename to `NESTED_ARCHIVE_REGEX`?

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
```diff
@@ -150,147 +151,146 @@ private static String extractArtefactUrl(URL jarOrClassFolderUrl) {
...
+	private static List<Pair<String, Properties>> findGitPropertiesInDirectoryFile(
+			File directoryFile, boolean recursiveSearch) throws IOException {
+		List<Pair<String, Properties>> result = new ArrayList<>();
+		result.addAll(findGitPropertiesInFolder(directoryFile));
 
-		} catch (IOException e) {
-			throw new IOException("Reading jar " + firstPart + " for obtaining commit " +
-					"descriptor from git.properties failed", e);
+		if (recursiveSearch) {
+			result.addAll(findInJarFilesNestedGitPropertiesFilesInFolder(directoryFile));
```
Raphael-N, line 217: Maybe you could split this function into `findJarFilesInFolder` and `findGitPropertiesInJarFiles` or something like that. The name makes me think that this function has too many responsibilities 😄

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocator.java
AnonymFx: When is `data.size() > 1` possible: In every case in which the given jar file or class folder contains multiple git.properties files. E.g. when you have nested jars with git.properties files.

Re inconsistent handling of errors: Yeah, you're right. We can just use the first one we found. 
Also, I assume with "here" in "here, we log an error" you mean some lines below. As we continue after the error, I'd change that to warning. I also changed the debug log for `data.size() > 1` to a warning.

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
AnonymFx: can other archives then jars be nested? I can only imagine nested jar libs. ear/war/aar are application archives and are probably not nested within other archives.

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.java
AnonymFx: Hm, yes and no. I renamed it, hopefully that makes things clearer.
I implemented it in a way so that `findGitPropertiesInDirectoryFile` is clearly understandable and we don't get more small utility functions with vaguely the same name (findXinY) It's confusing at it is already 😉

agent/src/main/java/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocator.java
```diff
@@ -30,21 +31,26 @@
 	private final DelayedUploader<T> uploader;
 	private final DataExtractor<T> dataExtractor;
 
-	public GitPropertiesLocator(DelayedUploader<T> uploader, DataExtractor<T> dataExtractor) {
```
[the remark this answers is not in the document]
AnonymFx: Done (FYI, I keep discussion thread open because when I do reviews, I like to look at the resolution of the previous comments and that difficult if I don't see them anymore when they are resolved)

PR #219, opened 2022-07-28
TS-30956: handle tests with trailing whitespace correctly

andipunz:
Addresses issue [TS-30956](https://cqse.atlassian.net/browse/TS-30956)

- [x] CHANGELOG.md updated


teamscale-client/src/main/java/com/teamscale/client/ITeamscaleService.java
```diff
@@ -105,7 +105,7 @@ Call<ResponseBody> uploadExternalReports(
 	);
 
 	/** Retrieve clustered impacted tests based on the given available tests and baseline timestamp. */
-	@PUT("api/v7.6.0/projects/{projectName}/impacted-tests?prioritization-strategy=CHEAP_ADDITIONAL_COVERAGE_PER_TIME")
+	@PUT("api/v8.0.0/projects/{projectName}/impacted-tests?prioritization-strategy=CHEAP_ADDITIONAL_COVERAGE_PER_TIME")
```
DreierF, line 108: As we switched to `CHEAP_ADDITIONAL_COVERAGE_PER_TIME` being the default value i the meantime we can remove the extra parameter here that we only added for backwards compatibility

teamscale-client/src/main/java/com/teamscale/client/ITeamscaleService.java
```diff
@@ -115,11 +115,11 @@ Call<List<PrioritizableTestCluster>> getImpactedTests(
 			@Query("include-failed-and-skipped") boolean includeFailedAndSkippedTests,
 			@Query("ensure-processed") boolean ensureProcessed,
 			@Query("include-added-tests") boolean includeAddedTests,
-			@Body List<ClusteredTestDetails> availableTests
+			@Body List<TestWithClusterId> availableTests
 	);
 
 	/** Retrieve unclustered impacted tests based on all tests known to Teamscale and the given baseline timestamp. */
-	@GET("api/v7.6.0/projects/{projectName}/impacted-tests?prioritization-strategy=CHEAP_ADDITIONAL_COVERAGE_PER_TIME")
+	@GET("api/v8.0.0/projects/{projectName}/impacted-tests?prioritization-strategy=CHEAP_ADDITIONAL_COVERAGE_PER_TIME")
```
DreierF, line 122: Same here

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ We use [semantic versioning](http://semver.org/):
 - PATCH version when you make backwards compatible bug fixes.
 
 # next release
+- [fix] _impacted-test-engine_: Handle tests with trailing whitespaces correctly
```
DreierF, line 8:
```suggestion
- [fix] _impacted-test-engine_: Handle tests with trailing whitespaces correctly
- [breaking] _impacted-test-engine_: Now requires Teamscale 8.0 as minimum version
```

Maybe we should also document this minimum supported version in the README of the gradle and maven plugins, which is what users typically interact with

teamscale-client/src/main/java/com/teamscale/client/ClusteredTestDetails.java
```diff
@@ -12,23 +12,17 @@ public class ClusteredTestDetails extends TestDetails {
...
-	}
-
 	/**
-	 * Creates clustered test details with the given additional {@link TestData}.
-	 * <p>
-	 * Use this to easily mark additional files or data as belonging to that test case. Whenever the given {@link
-	 * TestData} changes, this test will be selected to be run by the TIA.
-	 * <p>
-	 * Example: For a test that reads test data from an XML file, you should pass the contents of that XML file as its
-	 * test data. Then, whenever the XML is modified, the corresponding test will be run by the TIA.
```
DreierF, line 27: Why did we remove the documentation comment? As that is "public API" of our teamscale-client which is used at customers we should keep the documentation.

DreierF (changes requested):

teamscale-client/src/main/java/com/teamscale/client/ClusteredTestDetails.java
andipunz: I did not remove the documentation comment but a whole unused function. See class before the changes:
`public class ClusteredTestDetails extends TestDetails {

	/**
	 * A unique identifier for the cluster this test should be prioritized within. If null the test gets assigned it's
	 * own unique cluster.
	 */
	public String clusterId;

	public ClusteredTestDetails(String uniformPath, String sourcePath, String content, String clusterId) {
	public String partition;

	public ClusteredTestDetails(String uniformPath, String sourcePath, String content, String clusterId,
								String partition) {
		super(uniformPath, sourcePath, content);
		this.clusterId = clusterId;
	}

	/**
	 * Creates clustered test details with the given additional {@link TestData}.
	 * <p>
	 * Use this to easily mark additional files or data as belonging to that test case. Whenever the given {@link
	 * TestData} changes, this test will be selected to be run by the TIA.
	 * <p>
	 * Example: For a test that reads test data from an XML file, you should pass the contents of that XML file as its
	 * test data. Then, whenever the XML is modified, the corresponding test will be run by the TIA.
	 */
	public static ClusteredTestDetails createWithTestData(String uniformPath, String sourcePath, TestData testData,
														  String clusterId) {
		return new ClusteredTestDetails(uniformPath, sourcePath, testData.hash, clusterId);
		this.partition = partition;
	}

}`

DreierF: Actually I would tend to keep the method + doc comment and use it in our custom test framework test as it was added by Fabi AFAIR for that use case

PR #264, opened 2023-05-25
TS-8032 Maven External Analysis Upload

AlEscher:
Addresses issue [TS-8032](https://cqse.atlassian.net/browse/TS-8032)

- [x] CHANGELOG.md updated


AlEscher:
The newly added goal performs the following steps for a Maven project (and its subprojects):
1. For each Jacoco report type (report, report-integration, report-aggregate)
  - Check that the report goal is set to produce `XML` reports
  - Read the output directory, in case a custom one was specified by the user
2. Collect all reports and upload them to the Teamscale instance. All reports of one goal are uploaded in the same session and to the same partition

AnonymFx (changes requested):
Excellent work! I left a couple of comments inline. Also:
* I wanted to do quick test with pixelitor (attached) but it didn't work (our plugin found no reports). For some reason the XML report is under `target/site/jacoco/jacoco.xml` (notice the additional "site" in between). AFAICT this is not specified in the pom 🤔 Could you check what's up with that? [maven-upload-test.zip](https://github.com/cqse/teamscale-jacoco-agent/files/11616396/maven-upload-test.zip)
* Is there a ticket for writing a docs article already? We should definitely write docs for the feature 😉
* Please update CHANGELOG.md
* Do you have time to attend one of the next N&N (this week or the one afterwards)? Could you present this feature there?

AlEscher:
I have implemented most suggested changes, thank you very much for the review!
The attached project should now be working as well (I had to also change the surefire plugin version to 3.3.2).
Making a system test for a project without XML reports would be difficult since the test case invokes the plugin goal as an external process. The only way to detect that the incorrect configuration was detected would be via parsing the logs. If you think this would be ok @jveihelmann then I can still that implement that.
Regarding the documentation, where should this feature be documented?

system-tests/maven-external-upload-test/src/test/java/com/teamscale/upload/MavenExternalUploadSystemTest.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	private ProcessUtils.ExecutionResult runCoverageUploadGoal(String projectPath) {
+		File workingDirectory = new File(projectPath);
+		try {
+			return ProcessUtils.execute(new ProcessBuilder("./mvnw", "com.teamscale:teamscale-maven-plugin:upload-coverage").directory(workingDirectory));
+		} catch (IOException e) {
+			fail(String.valueOf(e));
+		}
+		return null;
```
jveihelmann, line 50: This method would probably fail on a windows machine (not in WSL) because of the `./mwnw` command. I would suggest to check the current os first and change the process builder command in the following way: 
```suggestion
        File workingDirectory = new File(projectPath);
	String executable = "./mvnw";
		if (SystemUtils.isWindows()) {
			executable = Paths.get(projectPath, "mvnw.cmd").toUri().getPath();
		}
		try {
	return ProcessUtils.execute(new ProcessBuilder(executable , MAVEN_COVERAGE_UPLOAD_GOAL).directory(workingDirectory));
	}
	catch (IOException e) {
			fail(String.valueOf(e));
		}
		return null;
```
I don't know why, but on my windows machine it was important to prefix the project path to the executable command, even when setting the `workingDirectory` with `.directory(workingDirectory)`

Also, please add a constant for "com.teamscale:teamscale-maven-plugin:upload-coverage"

teamscale-maven-plugin/src/main/java/com/teamscale/maven/TeamscaleMojoBase.java
```diff
@@ -0,0 +1,131 @@
...
+		Path basedir = session.getCurrentProject().getBasedir().toPath();
+		try {
+			GitCommit commit = GitCommit.getGitHeadCommitDescriptor(basedir);
+			if (StringUtils.isBlank(resolvedCommit)) {
+				resolvedCommit = commit.branch + ":" + commit.timestamp;
+			}
+			resolvedRevision = commit.sha1;
+		} catch (IOException e) {
+			throw new MojoFailureException("You did not configure an <endCommit> in the pom.xml" +
+					" and I could also not determine the checked out commit in " + basedir + " from Git", e);
```
jveihelmann, line 107: I think it's odd that this is written in the first person here: 
```suggestion
			throw new MojoFailureException("There is no <endCommit> configured in the pom.xml" +
					" and it was not possible to determine the checked out commit in " + basedir + " from Git", e);
```

teamscale-maven-plugin/src/main/java/com/teamscale/maven/TeamscaleMojoBase.java
```diff
@@ -0,0 +1,131 @@
...
+				return;
+			}
+		}
+		Path basedir = session.getCurrentProject().getBasedir().toPath();
+		try {
+			GitCommit commit = GitCommit.getGitHeadCommitDescriptor(basedir);
+			if (StringUtils.isBlank(resolvedCommit)) {
+				resolvedCommit = commit.branch + ":" + commit.timestamp;
+			}
+			resolvedRevision = commit.sha1;
```
jveihelmann, line 104: Now if a endCommit was explicitely set in the pom.xml but no revision, this would stil try to search for the commit in a git repository. So if there is no git repository, the Mojo would crash I think? 

I think we need to respect this because the gitRevision property did not exist in the original TiaMojoBase, so it is not essential. I would suggest to create a seperate method resolveRevision, that is only called from within the CoverageUploadMojo. WDYT?

teamscale-maven-plugin/src/main/java/com/teamscale/maven/upload/CoverageUploadMojo.java
```diff
@@ -0,0 +1,196 @@
...
+		uploadCoverage(reportAggregateGoalOutputFiles, aggregatedTestPartition);
+	}
+
+	private void uploadCoverage(List<Path> reportOutputFiles, String partition) throws IOException {
+		List<File> reports = new ArrayList<>();
+		getLog().debug(String.format("Scanning through %d locations for %s...", reportOutputFiles.size(), partition));
+		for (Path reportPath : reportOutputFiles) {
+			File report = reportPath.toFile();
+			if (!report.exists()) {
+				getLog().debug(String.format("Cannot find %s, skipping...", report.getAbsolutePath()));
```
jveihelmann, line 153:
```suggestion
				getLog().warn(String.format("Cannot find %s, skipping...", report.getAbsolutePath()));
```
I think this should be a warning aswell.

jveihelmann: Ah okay that makes sense, thanks for the explanation!

PR #361, opened 2023-11-08
TS-33237 Receive profiler configuration from Teamscale

DreierF:
Addresses issue [TS-33237](https://cqse.atlassian.net/browse/TS-33237)

- [x] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/configuration/ConfigurationViaTeamscale.java
```diff
@@ -0,0 +1,129 @@
...
+ * Responsible for holding the configuration that was retrieved from Teamscale and sending regular heartbeat events to
+ * keep the profiler information in Teamscale up to date.
+ */
+public class ConfigurationViaTeamscale {
+
+	/**
+	 * Two minute timeout. This is quite high to account for an eventual high load on the Teamscale server. This is a
+	 * tradeoff between fast application startup and potentially missing test coverage.
+	 */
+	private static final int LONG_TIMEOUT = 120;
```
stahlbauer, line 31: I would expect `java.time.Duration` to be used here.

agent/src/main/java/com/teamscale/jacoco/agent/configuration/ConfigurationViaTeamscale.java
```diff
@@ -0,0 +1,129 @@
...
+	private final String profilerId;
```
stahlbauer, line 32: What about also documenting this attribute?

agent/src/main/java/com/teamscale/jacoco/agent/configuration/ProcessInformationRetriever.java
```diff
@@ -0,0 +1,66 @@
...
+import com.teamscale.client.ProcessInformation;
+import com.teamscale.report.util.ILogger;
+
+import java.lang.management.ManagementFactory;
+import java.lang.reflect.InvocationTargetException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+/**
+ * The ProcessInformationRetriever class is responsible for retrieving process information such as the host name and
```
stahlbauer, line 12: Consider following the JavaDoc guideline, in particular `Class/interface/field descriptions can omit the subject and simply state the object.`.

agent/src/main/java/com/teamscale/jacoco/agent/configuration/ProcessInformationRetriever.java
```diff
@@ -0,0 +1,66 @@
...
+	/**
+	 * Retrieves the host name of the local machine.
+	 */
+	private String getHostName() {
+		try {
+			InetAddress inetAddress = InetAddress.getLocalHost();
+			return inetAddress.getHostName();
+		} catch (UnknownHostException e) {
+			logger.error("Failed to determine hostname!", e);
+			return "";
```
stahlbauer, line 41: What about alternatives such as the IP address itself?

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptionsParser.java
```diff
@@ -46,31 +48,32 @@ public class AgentOptionsParser {
 	private final ILogger logger;
 	private final FilePatternResolver filePatternResolver;
 	private final TeamscaleConfig teamscaleConfig;
-	private final String environmentConfigFile;
+	private final String environmentConfigId;
 	private final TeamscaleCredentials credentials;
 
-	public AgentOptionsParser(ILogger logger, String environmentConfigFile,
+	public AgentOptionsParser(ILogger logger, String environmentConfigId,
```
stahlbauer, line 54: What is the `environmentConfigId`. Add some documentation?

agent/src/main/java/com/teamscale/jacoco/agent/configuration/ProcessInformationRetriever.java
DreierF: AFAICS you'd retrieve the IP in Java via 
```
InetAddress ip = InetAddress.getLocalHost();
System.out.println("Your IP address is: " + ip.getHostAddress());
```
which is exactly the same function that we are currently catching the exception of. Do you see any alternative here?

stahlbauer: I would have expected that if you already have an ip address at hand (the `ip`) object, you could provided this information. I was assuming that the call to `getHostHame` could lead to an inverse lookup of the hostname as described in the docs ` reverse name lookup will be performed and the result will be returned based on the system configured name lookup service` in would have an exception in case this lookup fails. But I do not know the very details of the implementation here. See also https://stackoverflow.com/questions/7348711/recommended-way-to-get-hostname-in-java for fallback solutions.

DreierF: Which fallback solution are you referring to?

agent/MIGRATION.md
```diff
@@ -0,0 +1,24 @@
...
+```sh
+./installer install https://your.teamscale.url your-build-user-name your-access-token
+```
+
+This will install the agent to `/opt/teamscale-profiler/java` and will attach it to all Java applications.
+
+Then for each application where you currently use the Teamscale JaCoCo Agent: 
+- Create a profiler configuration within Teamscale with the same options that you previously specified via the `config-file=agent.properties`
+  - Omit the `teamscale-server-url`, `teamscale-user` and `teamscale-access-key` options
+- Instead, attach the `TEAMSCALE_JAVA_PROFILER_CONFIG_ID=my-configuration-id` environment variable to the start script of your System under Test
```
karottenreibe, line 20: `start script` may be confusing. E.g. if you run a Tomcat as a systemd service, that doesn't have a start script as such.

How about
> Instead, set the `TEAMSCALE_JAVA_PROFILER_CONFIG_ID=my-configuration-id` environment variable for the Java process that runs your system under test (e.g. the application server or systemd service).

agent/MIGRATION.md
```diff
@@ -0,0 +1,24 @@
...
+- Remove the existing configuration that manually attaches the Teamscale JaCoCo Agent to the System under Test  
+- Restart the application and verify that the profiler is shown in Teamscale under `Running Profiler`
```
karottenreibe, line 22: Running profiler**s**

agent/src/main/java/com/teamscale/jacoco/agent/util/AgentUtils.java
```diff
@@ -23,7 +23,7 @@ public static Path getMainTempDirectory() {
 		if (mainTempDirectory == null) {
 			try {
 				mainTempDirectory = Files.createTempDirectory("teamscale-java-profiler-" +
-						FileSystemUtils.toSafeFilename(ProcessInformationRetriever.getPID()) + "-");
+						FileSystemUtils.toSafeFilename(ProcessInformationRetriever.getPID()));
```
karottenreibe, line 26: Please leave the `-` at the end. It separates the PID from the random string that is appended by Java to make the directory unique. Being able to quickly see the PID from the directory name may be useful during debugging to identify the temp directory (containing e.g. the log files) of a problematic process!

With the dash you get directories like `/tmp/teamscale-java-profiler-1234-15687546546573`
Without they look like `/tmp/teamscale-java-profiler-123415687546546573` where it's not clear what the PID was for that directory.

karottenreibe: Maybe add a comment with this rationale so others don't do the same thing again in the future? :P

karottenreibe (changes requested):

agent/src/main/java/com/teamscale/jacoco/agent/util/AgentUtils.java
DreierF: 💡

agent/src/main/java/com/teamscale/jacoco/agent/configuration/ProcessInformationRetriever.java
stahlbauer: I was referring to the alternatives to get the host machines name. But we can wait if this problem manifests in the end and fix it on demand. So this is fine for me for now.

PR #363, opened 2023-11-14
TS-36294 Add new maven plugin mojo to batch convert .exec files to testwise coverage.

jveihelmann:
Addresses issue [TS-36294](https://cqse.atlassian.net/browse/TS-36294)

- [x] CHANGELOG.md updated


teamscale-maven-plugin/src/main/java/com/teamscale/tia/maven/TiaCoverageConvertMojo.java
```diff
@@ -0,0 +1,135 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Batch converts all created .exec file reports into a testwise coverage report.
+ */
+@Mojo(name = "testwise-coverage-converter", defaultPhase = LifecyclePhase.VERIFY,
```
AnonymFx, line 35: `defaultPhase = LifecyclePhase.VERIFY` means we don't do the conversion when executing `mvnw test` but only when calling `mvn verify`. Can we make it work for both out of the box?

teamscale-maven-plugin/src/main/java/com/teamscale/tia/maven/TiaCoverageConvertMojo.java
```diff
@@ -0,0 +1,135 @@
...
+		requiresDependencyResolution = ResolutionScope.RUNTIME, threadSafe = true)
```
AnonymFx, line 36: `threadSafe = true` are we actually? Have you tested that? What happens when two of these processes run with the same configuration? File conflicts?!

AnonymFx, line 36: Do you know what `requiresDependencyResolution = ResolutionScope.RUNTIME` is used for?

AnonymFx (changes requested):

teamscale-maven-plugin/src/main/java/com/teamscale/tia/maven/TiaCoverageConvertMojo.java
jveihelmann: Not sure, I tested this on the multiple process test and it seemed to create a correct testwise coverage json. I will set it to false just to make sure.

jveihelmann: I guess this only tells maven that the dependencies do not have to be resolved during compilation, but at runtime. I think it does not make a great difference, but I adapted this from our other plugin mojos.

teamscale-maven-plugin/src/main/java/com/teamscale/tia/maven/TiaCoverageConvertMojo.java
jveihelmann: This is intentional, as this makes sure that the plugin is executed after both the unit and the integration tests are run. Setting the default to `LifecyclePhase.TEST` would mean that the coverage is converted before the integration tests are executed.

If one desires to execute the plugin after running only mvnw test, you have to either explicitely execute the goal in the run command something like ->` mvnw test com.teamscale.tia.maven:testwise-coverage-converter`

Or by changing the `pom.xml` to execute the mojo at a specific phase like this:
```
                    <execution>
                        <id>test-report-conversion</id>                    
                        <goals>
                            <goal>testwise-coverage-converter</goal>
                        </goals>
                        <phase>test</phase>
                    </execution>
```

teamscale-maven-plugin/src/main/java/com/teamscale/tia/maven/TiaMojoBase.java
```diff
@@ -286,19 +286,7 @@ private void validateTestPluginConfiguration(PluginExecution execution) throws M
...
 		validateParallelizationParameter(configurationDom, "threadCount");
 		validateParallelizationParameter(configurationDom, "forkCount");
-
-		Xpp3Dom parameterDom = configurationDom.getChild("reuseForks");
-		if (parameterDom == null) {
-			return;
-		}
-
-		String value = parameterDom.getValue();
-		if (value != null && !value.equals("true")) {
```
AnonymFx, line 296: I know this does not work with all the ways you can configure reuse forks but I think we should keep the log **as a warning** regardless. We'll need to change the wording a bit, though. We shouldn't state that it's not working at all, but has some performance impact. Maybe you can reuse parts of the docs change :)

teamscale-maven-plugin/src/main/java/com/teamscale/tia/maven/TiaCoverageConvertMojo.java
AnonymFx: I see, thanks for the explanation.

AnonymFx (changes requested):
Additionally to the inline comment, please 
* resolve the merge conflict
* test, if/how the upload with the new upload goal works and add instructions to the docs (there's also a corresponding comment in the merge request on GitLab :)

teamscale-maven-plugin/src/main/java/com/teamscale/maven/upload/CoverageUploadMojo.java
```diff
@@ -101,53 +144,85 @@ public void execute() throws MojoFailureException {
...
+				.resolve(jacocoDirectory).resolve("jacoco.xml");
+		getLog().debug(String.format("Adding possible report location: %s", resolvedCoverageFile));
+		reportOutputFiles.add(resolvedCoverageFile);
 	}
 
 	private void uploadCoverageReports() throws IOException {
-		uploadCoverage(reportGoalOutputFiles, unitTestPartition);
-		uploadCoverage(reportIntegrationGoalOutputFiles, integrationTestPartition);
-		uploadCoverage(reportAggregateGoalOutputFiles, aggregatedTestPartition);
+		if (uploadTestwiseCoverage) {
```
AnonymFx, line 206: Why don't we just upload everything all the time as we do with unit, integration and aggregated test coverage? Do you have some specific use case in mind?

jveihelmann: I thought that it seems unlikely that someone uploads testwise coverage while also wanting to upload raw jacoco coverage. Can change it so that everything is uploaded all the time. WDYT?

AnonymFx (changes requested):

PR #401, opened 2024-01-18
TS-37029 Remove http upload

AlEscher:

- [x] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
```diff
@@ -378,10 +368,6 @@ private void validateTestwiseCoverageConfig(Validator validator) {
 				httpServerPort != null && testImpactConfig.testEnvironmentVariable != null,
 				"You did set both 'http-server-port' and 'test-env'! Only one of them is allowed!");
 
-		validator.isFalse(uploadUrl != null,
-				"'upload-url' option is " +
-						"incompatible with Testwise coverage mode!");
-
```
jveihelmann, line 384: So after reading this part, I was wondering if we need `additionalMetaDataFiles` at all anymore, but I realized that they are used in the contexdt of artifactory and azure uploads. However, this line suggests that they need to be used in combination with the `uploadUrl` which I found strange. Can you explain how this can work?

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
```diff
@@ -269,9 +262,6 @@ public String getObfuscatedOptionsString() {
 			validateLoggingConfig(validator);
 		}
 
-		validator.isFalse(uploadUrl == null && !additionalMetaDataFiles.isEmpty(),
```
AlEscher, line 272: See discussion in https://cqse.slack.com/archives/CL0REH3FW/p1705588368099629
The `additionalMetaDataFiles` are passed directly to e.g. the `ArtifactoryUploader` which relies on the URL from `ArtifactoryConfig`, which is a different option than the `upload-url`.
I am not sure about the reason for this check here, since i.m.o. `additionalMetaDataFiles` does not rely only on `uploadUrl` (maybe it did before other upload functionalities where added).

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
AlEscher: I think this comment is probably set on the wrong change (at least for me it is) ^^' 
I replied here: https://github.com/cqse/teamscale-jacoco-agent/pull/401/files#r1459138353

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptions.java
jveihelmann: Ah yes, thanks for moving the comment :) 

I see Fabis reply which suggests that we can delete the `upload-metadata` as well. Correct me if I am wrong, but this suggests that we can get rid of the whole additionalMetaDataFiles. I would wait for his reply and then get back. :)

jveihelmann: Okay I think we should keep it.

PR #425, opened 2024-03-17
read config file path from TEAMSCALE_JAVA_PROFILER_CONFIG_FILE

karottenreibe:
Addresses issue [TS-38049](https://cqse.atlassian.net/browse/TS-38049)

- [x] CHANGELOG.md updated
- [x] Present new features in [N&N](https://cqse.atlassian.net/l/cp/KHSr81m1)


agent/src/main/java/com/teamscale/jacoco/agent/PreMain.java
```diff
@@ -55,15 +58,16 @@ public static void premain(String options, Instrumentation instrumentation) thro
 		System.setProperty(LOCKING_SYSTEM_PROPERTY, "true");
 
 		String environmentConfigId = System.getenv(CONFIG_ID_ENVIRONMENT_VARIABLE);
-		if (StringUtils.isEmpty(options) && environmentConfigId == null) {
+		String environmentConfigFile = System.getenv(CONFIG_FILE_ENVIRONMENT_VARIABLE);
+		if (StringUtils.isEmpty(options) && environmentConfigId == null && environmentConfigFile == null) {
 			// profiler was registered globally and no config was set explicitly by the user, thus ignore this process
 			// and don't profile anything
 			return;
 		}
```
jveihelmann, line 67: I just tested locally what happens if you set both environment variables. On windows, it looks like the config-file variable takes priority, and the config from Teamscale declared with the ID will be ignored. While this is a very artificial problem (as we only need the config file variable if the config id does not work), it would be good to log something in case both are set. 
However, I can imagine this could be confusing, if a customer switches from the config file system to the config id system and we set the id variable and have to wonder why that does not work. WDYT?

jveihelmann: Perhaps this can be done in the AgentOptionsParser instead.

jveihelmann (commented):
Thanks for working on this! Looks great and worked for me as expected. 
 
 Just nitpick comments from my side :) 
 Also, I would say to tolerate the Teamscale finding.

PR #558, opened 2024-09-23
Ts/40412 add agent specific proxy settings options

flameJam:
Addresses issue [TS-40412](https://cqse.atlassian.net/browse/TS-40412)

- [ ] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptionsParser.java
```diff
@@ -156,6 +156,9 @@ private void handleOption(AgentOptions options,
 						value)) {
 			return;
 		}
+		if (key.startsWith("proxy-") && TeamscaleProxyOptions.handleTeamscaleProxyOptions(options.teamscaleProxyOptions, key, value, filePatternResolver)) {
```
flameJam, line 159: IMO its not necessary to resolve the finding regarding the length of this function. We don't tolerate findings regarding method length, so I would just ignore it.

teamscale-client/src/main/java/com/teamscale/client/HttpUtils.java
```diff
@@ -102,24 +102,34 @@ private static void setUpProxyServer(OkHttpClient.Builder httpClientBuilder) {
 
 	private static boolean setUpProxyServerForProtocol(ProxySystemProperties.Protocol protocol,
 													   OkHttpClient.Builder httpClientBuilder) {
+		TeamscaleProxySystemProperties teamscaleProxySystemProperties = new TeamscaleProxySystemProperties(protocol);
```
flameJam, line 105: Why is Teamscale fine with the dependency on `ProxySystemProperties` one line below? Not sure what to do with those findings. We could either move the TeamscaleProxySystemProperties` into the utils in the architecture or do a bigger refactoring. Or ignore it :man_shrugging:

jveihelmann: It looks like the client component in https://cqse.teamscale.io/architecture/editor/teamscale-jacoco-agent/teamscale-jacoco-agent.architecture?t=ts%2F37747_proxy_password_file_option%3AHEAD contains the ProxySystemProperties on the feature branch, but not in master. I have no idea why this architecture is the way it is. @DreierF Do you know something about this? I would assume we can ignore this finding.

DreierF: The intention of the architecture here is that the utils component should be used from the client, but utils should not depend on classes from within the client. Not sure where the difference in the architecture definition on both branches comes from, but on your branch the ProxySystemProperties seems to have been added to the Utils component, which "solves the issue". So you could also define that TeamscaleProxySystemProperties belongs to Utils on master (i.e. after the ticket is merged) to fix the issue.

teamscale-client/src/main/java/com/teamscale/client/ProxySystemProperties.java
```diff
@@ -51,120 +58,100 @@ public ProxySystemProperties(Protocol protocol) {
...
 	@NotNull
-	private String getProxyPasswordSystemPropertyName() {
-		return protocol + PROXY_PASSWORD_SYSTEM_PROPERTY;
+	protected String getProxyPasswordSystemPropertyName() {
+		return getPropertyPrefix() + protocol + PROXY_PASSWORD_SYSTEM_PROPERTY;
 	}
 
 	/** Parses the given port string. Returns -1 if the string is null or not a valid number. */
-	private int parsePort(String portString) {
+	private int parsePort(String portString, Consumer<String> logFunction) {
```
flameJam, line 146: To be honest I very much dislike this. The problem is that we need the `ProxySystemProperties`/`TeamscaleProxySystemProperties` in the `TeamscaleProxyOptions`, and at that point we would have to use a `DelayedLogger`, as it happens at runtime before our logging config has been read and the default one does not allow "shaded FQCNs". But later in time we want to be able to log to a real logger... Do you have any idea how to improve this? I'm open to having a brainstorming session.

agent/src/main/java/com/teamscale/jacoco/agent/options/TeamscaleProxyOptions.java
```diff
@@ -0,0 +1,115 @@
...
+			}
+			if (String.format("proxy-%s-user", protocol).equals(key)) {
+				proxyUser = value;
+				return true;
+			}
+			if (String.format("proxy-%s-password", protocol).equals(key)) {
+				proxyPassword = value;
+				return true;
+			}
+			return false;
```
jveihelmann, line 79: If you decide to use the stripped prefix that I suggested before, you need to update the strings here as well.

teamscale-client/src/test/java/com/teamscale/client/ProxySystemPropertiesTest.java
```diff
@@ -16,6 +18,7 @@ static void teardown() {
 	@Test
 	void testPortParsing() {
 		properties.setProxyPort(9876);
+		Consumer<String> logFunction = logMessage -> {};
```
jveihelmann, line 21: Can be removed :)

agent/src/main/java/com/teamscale/jacoco/agent/AgentBase.java
```diff
@@ -60,21 +60,14 @@ public AgentBase(AgentOptions options) throws IllegalStateException {
...
-
+	/**
+	 * Stores the agent options for proxies in the {@link TeamscaleProxySystemProperties} and overwrites
+	 * the password with the password found in the proxy-password-file if necessary.
+	 * public for testing.
+	 */
+	 public static void putTeamscaleProxyOptionsIntoSystemProperties(AgentOptions options) {
+		options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTP).putTeamscaleProxyOptionsIntoSystemProperties();
+		options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTPS).putTeamscaleProxyOptionsIntoSystemProperties();
 	}
```
jveihelmann, line 71: Can you add the `@VisibleForTesting` Annotation instead? 
```suggestion
	/**
	 * Stores the agent options for proxies in the {@link TeamscaleProxySystemProperties} and overwrites
	 * the password with the password found in the proxy-password-file if necessary.
	 */
	 @VisibleForTesting
	 public static void putTeamscaleProxyOptionsIntoSystemProperties(AgentOptions options) {
		options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTP).putTeamscaleProxyOptionsIntoSystemProperties();
		options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTPS).putTeamscaleProxyOptionsIntoSystemProperties();
	}
```

teamscale-client/src/main/java/com/teamscale/client/ProxySystemProperties.java
```diff
@@ -51,122 +56,108 @@ public ProxySystemProperties(Protocol protocol) {
...
 
+	/** @return the name of the system property specifying the proxy password. */
 	@NotNull
-	private String getProxyPasswordSystemPropertyName() {
-		return protocol + PROXY_PASSWORD_SYSTEM_PROPERTY;
+	protected String getProxyPasswordSystemPropertyName() {
+		return getPropertyPrefix() + protocol + PROXY_PASSWORD_SYSTEM_PROPERTY;
+	}
+
+	public static class IncorrectPortFormatException extends IllegalArgumentException {
```
jveihelmann, line 143: Please add a java comment here :)

agent/src/main/java/com/teamscale/jacoco/agent/options/TeamscaleProxyOptions.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/** Stores the teamscale-specific proxy settings as system properties to make them always available. */
+	public void putTeamscaleProxyOptionsIntoSystemProperties() {
+		TeamscaleProxySystemProperties teamscaleProxySystemProperties = new TeamscaleProxySystemProperties(protocol);
+		if (proxyHost != null && !proxyHost.isEmpty()) {
```
jveihelmann, line 85: We can just use StringUtils.isEmpty() for that
```suggestion
		if (!StringUtils.isEmpty(proxyHost)) {
```

agent/src/main/java/com/teamscale/jacoco/agent/options/TeamscaleProxyOptions.java
```diff
@@ -0,0 +1,115 @@
...
+			teamscaleProxySystemProperties.setProxyHost(proxyHost);
+		}
+		if (proxyPort > 0) {
+			teamscaleProxySystemProperties.setProxyPort(proxyPort);
+		}
+		if(proxyUser != null && !proxyUser.isEmpty()) {
```
jveihelmann, line 91: Same as above

agent/src/main/java/com/teamscale/jacoco/agent/options/TeamscaleProxyOptions.java
```diff
@@ -0,0 +1,115 @@
...
+			teamscaleProxySystemProperties.setProxyUser(proxyUser);
+		}
+		if(proxyPassword != null && !proxyPassword.isEmpty()) {
```
jveihelmann, line 94: Same as above

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptionsParser.java
```diff
@@ -156,12 +163,39 @@ private void handleOption(AgentOptions options,
...
+		if (key.startsWith(httpPrefix)
+				&& options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTP).handleTeamscaleProxyOptions(StringUtils.stripPrefix(
+				key, httpPrefix), value)) {
+			return true;
+		}
+
+		if(key.equals("password-file")) {
+			Path proxyPasswordPath = filePatternResolver.parsePath(key, value);
+			options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTPS).proxyPasswordPath=proxyPasswordPath;
+			options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTP).proxyPasswordPath=proxyPasswordPath;
```
jveihelmann, line 193: Sorry I missed this before. Please use  a setter method for this instead of setting it with the raw value here. 
```suggestion
			Path proxyPasswordPath = filePatternResolver.parsePath(key, value);
			options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTPS).setProxyPasswordPath(proxyPasswordPath);
			options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTP).setProxyPasswordPath(proxyPasswordPath);
```

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptionsParser.java
```diff
@@ -156,12 +163,39 @@ private void handleOption(AgentOptions options,
...
+				key, httpsPrefix), value)) {
+			return true;
+		}
+
+		String httpPrefix = String.format("%s-", ProxySystemProperties.Protocol.HTTP);
+		if (key.startsWith(httpPrefix)
+				&& options.getTeamscaleProxyOptions(ProxySystemProperties.Protocol.HTTP).handleTeamscaleProxyOptions(StringUtils.stripPrefix(
+				key, httpPrefix), value)) {
+			return true;
+		}
```
jveihelmann, line 188: The way it is written now does not match what we tell the user to use as a parameter in the Readme. 

I just tested it locally and set the config parameter as: 
`proxy-host=localhost:80` but this fails, as the expected parameter would be
`http-proxy-host`. 

Please either update the agent/README.md or change the parameter, so that setting a `proxy-host` just sets it for both http and https.

agent/README.md
```diff
@@ -87,6 +88,20 @@ Note that defining `-Djava.io.tmpdir` will change the temp directory that is bei
...
+JVM.
+
+It is also possible to specify these options by prefixing the JVM flags for proxies with `teamscale.`,
+for example `-Dteamscale.https.proxyHost`.
+
+The following options are available:
+- `proxy-host`: The host name of the proxy server.
+- `proxy-port`: The port of the proxy server.
+- `proxy-user`: The username for the proxy server.
+- `proxy-password`: The password for the proxy user.
```
jveihelmann, line 103: I adressed this in my other commend. The option names do not match what we check for.

agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptionsParser.java
jveihelmann: The more I think about it the better I like the fact that we just keep the option as `proxy-host` and set it for both http and https.

flameJam: Tbh I'm not sure I agree. This would "mix" the two different behaviors of the global properties and the teamscale-specific ones, because the current implementation allows to use the global ones as defaults *for a specific protocol*, which are *selectively* overwritten by the Teamscale-specific options.

TLDR I'm not sure if increasing the complexity of the options with separate http-/https-options or the increasing the complexity by differing from the default behavior of the global options is more confusing to the users. 

WDYT?

jveihelmann: Okay you made a compelling case. Then we should just update the docs so the user knows what to type. ;)

PR #655, opened 2025-02-19
ts-41920 move config id validation to ignore order of options

jveihelmann:
Addresses issue [TS-41920](https://cqse.atlassian.net/browse/TS-41920)

- [x] CHANGELOG.md updated

### DIFF
--- a/agent/src/main/java/com/teamscale/jacoco/agent/PreMain.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/PreMain.java
@@ -1,8 +1,10 @@
 package com.teamscale.jacoco.agent;
 
-import com.teamscale.client.HttpUtils;
 import com.teamscale.jacoco.agent.configuration.AgentOptionReceiveException;
+import com.teamscale.jacoco.agent.logging.DebugLogDirectoryPropertyDefiner;
+import com.teamscale.jacoco.agent.logging.LogDirectoryPropertyDefiner;
 import com.teamscale.jacoco.agent.logging.LogToTeamscaleAppender;
+import com.teamscale.jacoco.agent.logging.LoggingUtils;
 import com.teamscale.jacoco.agent.options.AgentOptionParseException;
 import com.teamscale.jacoco.agent.options.AgentOptions;
 import com.teamscale.jacoco.agent.options.AgentOptionsParser;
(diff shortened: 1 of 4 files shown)


agent/src/main/java/com/teamscale/jacoco/agent/PreMain.java
```diff
@@ -131,7 +130,8 @@ private static Pair<AgentOptions, List<Exception>> getAndApplyAgentOptions(Strin
 
 		TeamscaleCredentials credentials = TeamscalePropertiesUtils.parseCredentials();
 		if (credentials == null) {
-			delayedLogger.warn("Did not find a teamscale.properties file!");
+			// With the current config mechanisms, this will almost always be shown.
+			delayedLogger.debug("No explicit teamscale.properties file given.");
```
jveihelmann, line 134: It's annoying to read this line in every log and can be confusing for users that don't know that we actually differentiate between a teamscale.properties file and config files (so they think the config file that they created is not found). 
I am not sure if we should get rid of this error alltogether or "hide" it as a debug log, so I chose the "less-extrem" variant.

agent/src/main/java/com/teamscale/jacoco/agent/PreMain.java
```diff
@@ -160,7 +160,6 @@ private static Pair<AgentOptions, List<Exception>> getAndApplyAgentOptions(Strin
 		initializeLogging(agentOptions, delayedLogger);
 		Logger logger = LoggingUtils.getLogger(Agent.class);
 		delayedLogger.logTo(logger);
-		HttpUtils.setShouldValidateSsl(agentOptions.shouldValidateSsl());
```
jveihelmann, line 163: Moved to AgentOptionsParser, as we need this before reaching out to Teamscale (see comment in AgentOptionsParser)

agent/src/main/java/com/teamscale/jacoco/agent/PreMain.java
```diff
@@ -131,7 +130,8 @@ private static Pair<AgentOptions, List<Exception>> getAndApplyAgentOptions(Strin
 
 		TeamscaleCredentials credentials = TeamscalePropertiesUtils.parseCredentials();
 		if (credentials == null) {
-			delayedLogger.warn("Did not find a teamscale.properties file!");
+			// With the current config mechanisms, this will almost always be shown.
```
DreierF, line 133: With `current` you mean the "old"/non-installer based setup right? This is IMHO confusing and will be even more confusing in the future. Maybe you can phrase this a bit clearer.

jveihelmann: Thanks for the comment. 
I updated it, wdyt?

DreierF: Way clearer 👍

PR #665, opened 2025-02-26
TS-41945 Support for configuration cache

DreierF:
Addresses issue [TS-41945](https://cqse.atlassian.net/browse/TS-41945)

- [x] CHANGELOG.md updated


teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
+# Migration guide for the `com.teamscale` Gradle plugin version 34 to 35
+Previously, the plugin provided a way to specify the partitions either globally for a project or on a per task level.
```
karottenreibe, line 2: It would be great if you not only mention what the previous state was but also what the new situation looks like.

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+Additionally, a `teamscaleReportUpload` was provided automatically on the root project that would lazily pick up all the reports and upload them to Teamscale.
```
karottenreibe, line 3:
```suggestion
Additionally, a `teamscaleReportUpload` task was provided automatically on the root project that would lazily pick up all the reports and upload them to Teamscale.
```

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+
+Due to the restrictions that are imposed by the Gradle configuration cache, we had to adjust the API of the plugin to provide compatibility with it.
```
karottenreibe, line 5:
```suggestion
With this version, we make use of the Gradle configuration cache for a faster configuration phase. Due to the restrictions that are imposed by the Gradle configuration cache, we had to adjust the API of the plugin.
```

Make it clearer why we even want to do this.

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+## Usage of `Property` for all inputs
+We switched all extension and task properties from plain types e.g. `boolean` to `Property<Boolean>` etc.
+In newer Gradle versions this should be handled transparently.
+If you are still using an older version, you might need to replace `partition = "Something"` with `partition.set("Something")`.
```
karottenreibe, line 11: do you know how old? otherwise it remains mysterious to the reader if they are affected or not ;-)

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+    // If you don't collect testwise coverage
+    from(tasks.test) // Will upload the JUnit reports produced by the Test task
+    from(tasks.jacocoTestReport) // Will upload the report produced by the JacocoReport task
+    
+    // or if you do collect testwise coverage
+    from(tasks.tiaTests) // Will upload the Testwise Coverage report produced by the TestImpacted task
+}
+```
+
+TODO aggregate at top-level
```
karottenreibe, line 35: todo left in the text

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+```groovy
+import com.teamscale.TeamscaleUpload
+// ...
+
+tasks.register("teamscaleReportUpload", TeamscaleUpload) {
+    partition = "Unit Tests"
+    message = "Gradle upload" // Optional
+    // If you don't collect testwise coverage
+    from(tasks.test) // Will upload the JUnit reports produced by the Test task
+    from(tasks.jacocoTestReport) // Will upload the report produced by the JacocoReport task
```
karottenreibe, line 28:
```suggestion
    from(tasks.jacocoTestReport) // Will upload the coverage report produced by the JacocoReport task
```

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+    
+    // or if you do collect testwise coverage
+    from(tasks.tiaTests) // Will upload the Testwise Coverage report produced by the TestImpacted task
+}
+```
+
+TODO aggregate at top-level
+
+## Specifying partitions top-level
+The `teamscale.report` extension action was removed.
```
karottenreibe, line 38:
```suggestion
The `teamscale.report` extension action was removed.
This was previously used to specify the partition to use during the report upload.
```

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+The partition is now specified via the `TeamscaleUpload` task.
```
karottenreibe, line 39:
```suggestion
The partition is now specified via the `TeamscaleUpload` task, see above.
```

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+with one of the following:
+If you have multiple `TestImpacted` tasks that should receive the same partition:
+```groovy
+import com.teamscale.TestImpacted
+// ...
+tasks.withType(TestImpacted) {
+    partition = "Unit Tests"
+}
+```
+If there is only one `TestImpacted` task to which the partition should apply:
```
karottenreibe, line 60: Can't I just also do the above in that case? That would simplify the instructions.

Additionally: why do I need to still specify the partition in the TestImpacted task? Shouldn't it be enough to specify it in the upload task? If not, then it's not clear from the text when to specify the partition in which location and why.

karottenreibe: Also, why can I only do this for the TestImpacted task but not for JUnit results or Jacoco coverage? seems inconsintent

teamscale-gradle-plugin/MIGRATION_35.md
```diff
@@ -0,0 +1,80 @@
...
+```groovy
+import com.teamscale.TestImpacted
+// ...
+tasks.register("tiaUnitTests", TestImpacted) {
+    partition = "Unit Tests"
+}
+```
+
+## `TestwiseCoverageReportTask` was merged into `TestImpacted`
+The testwise coverage report is now generated directly by the corresponding `TestImpacted` task (same like the native Gradle reports `junitXml` and `html`).
```
karottenreibe, line 70:
```suggestion
The testwise coverage report is now generated directly by the corresponding `TestImpacted` task (same like the native Gradle test result reports `junitXml` and `html`).
```

system-tests/gradle-multi-module/gradle-project/buildSrc/src/main/kotlin/com.example.java-convention.gradle.kts
```diff
@@ -12,39 +14,59 @@ java {
...
 	}
+	testClassesDirs = files(test.map { it.sources.output.classesDirs })
+	classpath = files(test.map { it.sources.runtimeClasspath })
+}
+
+val testwiseCoverageReportElements by configurations.creating {
+    isCanBeConsumed = true
+    isCanBeResolved = false
+	extendsFrom(configurations.getByName("implementation"))
+    attributes {
```
karottenreibe, line 67: indentation is off in this block

system-tests/gradle-multi-module/gradle-project/buildSrc/src/main/kotlin/com.example.aggregate.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+    attributes {
+		attribute(Category.CATEGORY_ATTRIBUTE, objects.named(Category.VERIFICATION))
+		attribute(Usage.USAGE_ATTRIBUTE, objects.named("testwise-coverage"))
+    }
+}
+
+tasks.register<TeamscaleUpload>("teamscaleReportUpload") {
+	partition = "Unit Tests"
+	from(tasks.named("tiaTests"))
+	addReport("TESTWISE_COVERAGE", testwiseCoverageReports.incoming.artifactView { lenient(true) }.files)
```
karottenreibe, line 21: in the migration guide this `addReport` is never mentioned. Reading the guide I would not have expected this to be necessary. I would have thought that the `from` call above already takes care of everything. This line is also rather cryptic with the `artifactview{lenient}` stuff. I don't understand what that is for

teamscale-gradle-plugin/MIGRATION_35.md
DreierF: Good point! I was in a hurry when I wrote this so didn't take the time to look up the details. Actually Groovy always supported this and Gradle 8.2 also enabled this by default for Kolin DSL. As we only support Gradle 8.4+ from now on, I just dropped that part.

teamscale-gradle-plugin/MIGRATION_35.md
DreierF: Because for the TestImpacted task we also need the partition to retrieve impacted tests from Teamscale. That's why the JUnit task for example does not need this.

system-tests/gradle-multi-module/gradle-project/buildSrc/src/main/kotlin/com.example.aggregate.gradle.kts
DreierF: This was the TODO part in the migration guide that I left out for now, as I was/am still hoping to find a less cryptic way of wiring those things up across projects. But as Gradle is going to forbid reaching into other projects e.g. to get a reference to their test tasks, the only clean way to do this is by exposing the reports as outgoing variants (consumable configurations) and retrieve them in other projects by declaring dependencies on them via a resolvable configuration. This is also how Gradle implements this natively in their [JaCoCo Report Aggregation Plugin](https://docs.gradle.org/current/userguide/jacoco_report_aggregation_plugin.html) and [Test Report Aggregation Plugin](https://docs.gradle.org/current/userguide/test_report_aggregation_plugin.html), but I still have to look closer into the implementation detail as to how we can achieve something similar without relying on users of the plugin using the [JVM Test Suite Plugin](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#jvm_test_suite_plugin) to organize their tests as I assume that the majority does not use this yet. Also [there is no way currently to make this work](https://github.com/gradle/gradle/blob/6b07f2944feb3aa974423c92e74968679538ec79/platforms/jvm/plugins-jvm-test-suite/src/main/java/org/gradle/api/plugins/jvm/internal/DefaultJvmTestSuiteTarget.java#L40) with `TestImpacted` tasks.

PR #670, opened 2025-03-07
improve error messages when configuration via Teamscale fails

karottenreibe:
Addresses issue [TS-42129](https://cqse.atlassian.net/browse/TS-42129)

- [x] CHANGELOG.md updated
- [x] Present new features in [N&N](https://cqse.atlassian.net/l/cp/KHSr81m1)

### DIFF
--- a/agent/src/main/java/com/teamscale/jacoco/agent/configuration/ConfigurationViaTeamscale.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/configuration/ConfigurationViaTeamscale.java
@@ -6,8 +6,8 @@ import com.teamscale.client.ProfilerConfiguration;
 import com.teamscale.client.ProfilerInfo;
 import com.teamscale.client.ProfilerRegistration;
 import com.teamscale.client.TeamscaleServiceGenerator;
-import com.teamscale.jacoco.agent.options.AgentOptionParseException;
 import com.teamscale.jacoco.agent.logging.LoggingUtils;
+import com.teamscale.jacoco.agent.options.AgentOptionParseException;
 import com.teamscale.report.util.ILogger;
 import okhttp3.HttpUrl;
 import okhttp3.ResponseBody;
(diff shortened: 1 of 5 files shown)


agent/src/main/java/com/teamscale/jacoco/agent/configuration/ConfigurationViaTeamscale.java
```diff
@@ -65,21 +65,21 @@ public static ConfigurationViaTeamscale retrieve(ILogger logger, String configur
...
 				throw new AgentOptionReceiveException(
-						"Failed to retrieve profiler configuration from Teamscale! " + response.errorBody().string());
+						"Failed to retrieve profiler configuration from Teamscale due to failed request. Http status: " + response.code()
+								+ " Body: " + response.errorBody().string());
 			}
 			ProfilerRegistration registration = response.body();
 			if (registration == null) {
-				throw new AgentOptionReceiveException("Failed to retrieve profiler configuration from Teamscale!");
+				throw new AgentOptionReceiveException(
+						"Failed to retrieve profiler configuration from Teamscale due to empty response. Status: " + response.code());
```
DreierF, line 75: I was wondering if we should include  `response.raw().body().string()` in the message. From the backend side I don't see how this could ever be `null`. So maybe Teamscale or a proxy etc. returns stuff here that just can't be deserialized correctly.

karottenreibe: I thought about that too, but by the time we catch the exception, the body will have been already consumed by the json parser, meaning this will give us a "closed" exception, right?

karottenreibe: AFAIU the body can only ever be read once :-( and we have no control over the json parsing doing so, since this happens automatically inside retrofit. So by the time we even get a response object, the body will already be gone

karottenreibe: good idea, implemented this :-) better be paranoid than sorry ;-)

PR #705, opened 2025-05-20
[TS-38628] [System Tests] Kotlin Migration New

Avanatiker:
Migrated System Tests to Kotlin


teamscale-client/src/main/kotlin/com/teamscale/client/JsonUtils.kt
```diff
@@ -52,18 +60,18 @@ object JsonUtils {
...
 		)
 
 	/**
 	 * Serializes an object into its JSON representation.
 	 */
 	@JvmStatic
 	@Throws(JsonProcessingException::class)
-	fun serialize(value: Any): String =
-		OBJECT_MAPPER.writeValueAsString(value)
+	fun Any.serialize(): String =
```
DreierF, line 73: Now that it is an extension function the name might be a bit too unspecific.

```suggestion
	fun Any.serializeToJson(): String =
```

PR #711, opened 2025-05-28
TS-42917 Ensure partition and clusterId are not null

snowbagoly:

- [ ] CHANGELOG.md updated

### DIFF
--- a/agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestEventHandlerStrategyBase.java
+++ b/agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestEventHandlerStrategyBase.java
@@ -128,7 +128,7 @@ public abstract class TestEventHandlerStrategyBase {
 		String partition = agentOptions.getTeamscaleServerOptions().partition;
 		if (partition == null) {
 			throw new UnsupportedOperationException(
-					"You must provide a partition via the agent's '" + TeamscaleConfig.TEAMSCALE_PARTITION_OPTION + "' option.");
+					"You must provide a partition via the agent's '" + TeamscaleConfig.TEAMSCALE_PARTITION_OPTION + "' option or using the /partition REST endpoint.");
 		}
 		return partition;
 	}


impacted-test-engine/src/main/kotlin/com/teamscale/test_impacted/engine/ImpactedTestEngine.kt
```diff
@@ -21,6 +21,10 @@ class ImpactedTestEngine : TestEngine {
 		if (!engineOptions.enabled) {
 			return EngineDescriptor(uniqueId, ENGINE_NAME)
 		}
+		if (engineOptions.partition == null) {
+			throw AssertionError("Agent option partition is undefined, but it's a mandatory parameter when executing impacted tests.")
+		}
```
DreierF, line 26:
```suggestion
		check(engineOptions.partition != null) { "Agent option partition is undefined, but it's a mandatory parameter when executing impacted tests." }
```

teamscale-gradle-plugin/src/main/kotlin/com/teamscale/TestImpactConfigurationAction.kt
```diff
@@ -92,6 +91,7 @@ class TestImpactConfigurationAction(
 		writeProperty("runAllTests", extension.runAllTests.get())
 		writeProperty("includeAddedTests", extension.includeAddedTests.get())
 		writeProperty("includeFailedAndSkipped", extension.includeFailedAndSkipped.get())
+		writeProperty("partition", extension.partition.get())
```
DreierF, line 94: We should only need to pass a partition here if we talk to Teamscale, which we don't do if runImpacted is false, so a partition might not be set here in all cases.

DreierF (changes requested):

agent/src/main/java/com/teamscale/jacoco/agent/testimpact/TestEventHandlerStrategyBase.java
```diff
@@ -114,6 +121,18 @@ public List<PrioritizableTestCluster> testRunStart(List<ClusteredTestDetails> av
...
 	}
 
+	/**
+	 * Returns the partition defined in the agent options. Asserts that the partition is defined.
+	 */
+	private String getPartition() {
+		String partition = agentOptions.getTeamscaleServerOptions().partition;
+		if (partition == null) {
+			throw new UnsupportedOperationException(
+					"You must provide a partition via the agent's '" + TeamscaleConfig.TEAMSCALE_PARTITION_OPTION + "' option.");
```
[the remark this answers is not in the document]
snowbagoly: Added it to the message now 😁

PR #855, opened 2026-02-24
[TS-38628] [Agent] Kotlin Migration

Avanatiker:
Migrating the agent to kotlin


agent/src/main/kotlin/com/teamscale/jacoco/agent/upload/azure/AzureFileStorageHttpUtils.kt
```diff
@@ -0,0 +1,117 @@
...
+			CONTENT_ENCODING, CONTENT_LANGUAGE, CONTENT_LENGTH, CONTENT_MD_5, CONTENT_TYPE, DATE, IF_MODIFIED_SINCE,
+			IF_MATCH, IF_NONE_MATCH, IF_UNMODIFIED_SINCE, RANGE
+		).map { headers.getOrDefault(it, "") }
+
+		val xmsHeader = headers.filter { it.key.startsWith("x-ms") }
+		return listOf(
+			httpMethod.toString(),
+			keys, xmsHeader.createCanonicalizedString(),
+			createCanonicalizedResources(account, path, queryParameters)
+		).joinToString()
```
DreierF: Those headers were joined using `String.join("\n"` in Java, but `joinToString` in Kotlin joins with `", "` by default

agent/src/main/kotlin/com/teamscale/jacoco/agent/convert/ConvertCommand.kt
```diff
@@ -0,0 +1,151 @@
...
+				"coverage or jacoco coverage should be generated."
+	)
+	var shouldGenerateTestwiseCoverage = false
+
+	/** After how many tests testwise coverage should be split into multiple reports.  */
+	@Parameter(
+		names = ["--split-after", "-s"], required = false, arity = 1, description = "After how many tests " +
+				"testwise coverage should be split into multiple reports (Default is 5000)."
+	)
+	val splitAfter = 5000
```
DreierF: Shouldn't this be `var` instead of `val` so that it can be modified via reflection? (We might want to switch JCommander out in a follow-up with a Kotlin friendly library)

agent/src/main/kotlin/com/teamscale/jacoco/agent/Main.kt
```diff
@@ -0,0 +1,81 @@
...
+	private fun createJCommanderBuilder() =
+		JCommander.newBuilder().programName(Main::class.java.getName())
+			.addObject(defaultArguments)
+			.addObject(command)
+
+	/** Default arguments that may always be provided.  */
+	private class DefaultArguments {
+		/** Shows the help message.  */
+		@Parameter(names = ["--help"], help = true, description = "Shows all available command line arguments.")
+		val help = false
```
DreierF: Maybe same problem here?

agent/src/main/kotlin/com/teamscale/jacoco/agent/PreMain.kt
```diff
@@ -0,0 +1,331 @@
...
+
+	/** Environment variable from which to read the Teamscale access token.  */
+	private const val ACCESS_TOKEN_ENVIRONMENT_VARIABLE = "TEAMSCALE_ACCESS_TOKEN"
+
+	/**
+	 * Entry point for the agent, called by the JVM.
+	 */
+	@JvmStatic
+	@Throws(Exception::class)
+	fun premain(options: String, instrumentation: Instrumentation?) {
```
DreierF: According to https://stackoverflow.com/a/36383986 `options` should also be nullable

agent/src/main/kotlin/com/teamscale/jacoco/agent/commandline/Validator.kt
```diff
@@ -0,0 +1,56 @@
...
+	private val messages = mutableListOf<String>()
+
+	/** Runs the given validation routine.  */
+	fun ensure(validation: ExceptionBasedValidation) {
+		try {
+			validation.validate()
+		} catch (e: Exception) {
+			e.message?.let { messages.add(it) }
+		} catch (e: AssertionError) {
+			e.message?.let { messages.add(it) }
```
DreierF: Could it be a problem that we skip over errors with `null` messages? `isValid` would return `true` now, `false` in Java

agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatingTransformer.kt
```diff
@@ -0,0 +1,78 @@
...
+class GitPropertiesLocatingTransformer(
+	private val locator: IGitPropertiesLocator,
+	private val locationIncludeFilter: ClasspathWildcardIncludeFilter
+) : ClassFileTransformer {
+	private val logger = getLogger(this)
+	private val seenJars = ConcurrentSkipListSet<String>()
+
+	override fun transform(
+		classLoader: ClassLoader?,
+		className: String,
```
DreierF: Not 100% sure if className is really non-nullable. It's hard to find any prove in either direction, but keeping it nullable would keep us on the safe side. Same holds for the other tree transformers.

agent/src/main/kotlin/com/teamscale/jacoco/agent/upload/delay/DelayedUploader.kt
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+
+	private fun uploadCachedXmls() {
+		try {
+			if (!cacheDir.isDirectory()) {
+				// Found data before XML was dumped
+				return
+			}
+			val xmlFilesStream = cacheDir.walk().filter { path ->
```
DreierF: Walk is recursive while the previous implementation use `listFiles`, i.e., non-recursive.

agent/src/main/kotlin/com/teamscale/jacoco/agent/PreMain.kt
```diff
@@ -0,0 +1,331 @@
...
+			LoggingUtils.loggerContext.getLogger(PreMain::class.java).error(e.message, e)
+
+			// Flush logs to Teamscale, if configured.
+			closeLoggingResources()
+
+			// Unregister the profiler from Teamscale.
+			agentOptions?.configurationViaTeamscale?.unregisterProfiler()
+
+			// Configuration errors must be visible to the user; let the JVM report them.
+			throw e
```
DreierF: Configuration problems like these are no longer re-thrown on master, but here the application would still crash.

agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.kt
```diff
@@ -0,0 +1,508 @@
...
+	 * Nested jar files will also be searched recursively if specified.
+	 */
+	@JvmStatic
+	@JvmOverloads
+	@Throws(IOException::class)
+	fun findGitPropertiesInArchive(
+		inputStream: JarInputStream,
+		archiveName: String?,
+		recursiveSearch: Boolean,
+		isRootArchive: Boolean = true
```
DreierF, line 333: Out of curiosity: Why was this needed? The Java version didn't have that parameter.

agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.kt
```diff
@@ -0,0 +1,508 @@
...
+		InvocationTargetException::class
+	)
+	private fun getVfsContentFolder(
+		jarOrClassFolderUrl: URL
+	): Pair<File, Boolean> {
+		// we obtain the URL of a specific class file as input, e.g.,
+		// vfs:/content/helloworld.war/WEB-INF/classes
+		// Next, we try to extract the artefact URL from it, e.g., vfs:/content/helloworld.war
+		val artefactUrl = extractArtefactUrl(jarOrClassFolderUrl)
+		val virtualFile = URI.create(artefactUrl).toURL().openConnection().getContent()
```
DreierF: `URI.create` is stricter about percent-encoding compared to `new URL` and JBoss VFS-style URLs with literal spaces may now throw.

agent/src/main/kotlin/com/teamscale/jacoco/agent/AgentBase.kt
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Base class for agent implementations. Handles logger shutdown, store creation and instantiation of the
+ * [JacocoRuntimeController].
+ *
+ *
+ * Subclasses must handle dumping onto disk and uploading via the configured uploader.
+ */
+abstract class AgentBase(
+	/** The agent options.  */
+	@JvmField var options: AgentOptions
```
DreierF: Why does this need `@JvmField`? Same for `controller`?

agent/src/main/kotlin/com/teamscale/jacoco/agent/testimpact/TestwiseCoverageResource.kt
```diff
@@ -0,0 +1,98 @@
...
+	companion object {
+		/** Path parameter placeholder used in the HTTP requests.  */
+		private const val TEST_ID_PARAMETER = "testId"
+
+		private var testwiseCoverageAgent: TestwiseCoverageAgent? = null
+
+		/**
+		 * Static setter to inject the [TestwiseCoverageAgent] to the resource.
+		 */
+		@JvmStatic
```
DreierF: Do we need this?

agent/src/main/kotlin/com/teamscale/jacoco/agent/options/AgentOptionsParser.kt
```diff
@@ -0,0 +1,540 @@
...
+			else -> return false
+		}
+	}
+
+	@Throws(AgentOptionParseException::class, AgentOptionReceiveException::class)
+	private fun readConfigFromTeamscale(options: AgentOptions) {
+		if (options.teamscaleServer.configId == null) return
+		if (!options.teamscaleServer.isConfiguredForServerConnection) {
+			throw AgentOptionParseException(
+				"Config-id '${options.teamscaleServer.configId}' specified without teamscale url/user/accessKey! These options must be provided locally via config-file or command line argument."
```
DreierF: The error message was much more specific on master compared to this version here. https://github.com/cqse/teamscale-java-profiler/blob/e822a567a410271e09922195fe7096cd3803430b/agent/src/main/java/com/teamscale/jacoco/agent/options/AgentOptionsParser.java#L424

DreierF (changes requested):

agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.kt
Avanatiker: As I reworked the whole function i thought it would make sense to treat the case of a nested empty archive as no exception. Thats what the parameter does. i can remove it if you want.

agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatorUtils.kt
Avanatiker: URL constructors are deprecated. I changed it now to a URI constructor that works with spaces.

agent/src/main/kotlin/com/teamscale/jacoco/agent/AgentBase.kt
Avanatiker: We actually still have a lot of those annotations sprinkled around in the codebase where most should not be necessary anymore. they are just artifacts from the migration where we needed intercompat with java in the process. ill remove them.

DreierF (approved):

PR #881, opened 2026-04-08
TS-45056 Class hot reloading detected and coverage dump fixed 

stahlbauer:
Addresses issue [TS-45056](https://cqse.atlassian.net/browse/TS-45056)

- [x] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -40,15 +44,18 @@
...
 	private Timer timer;
 
 	/** Stores the XML files. */
 	protected final IUploader uploader;
 
+	/**
+	 * Tracks class IDs (CRC64) by class name across dumps to detect application server reloads. When a known class
+	 * name appears with a different ID, we know the class was reloaded with different bytecode.
+	 */
+	private final Map<String, Long> previousClassIds = new HashMap<>();
```
DreierF: I feel like the agent itself is the wrong level of abstraction at where to track this

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -198,4 +208,40 @@ private void dumpReportUnsafe() {
...
+	 *
+	 * <p>This correctly distinguishes reloads from normal multi-classloader setups: in a multi-classloader setup,
+	 * both class IDs are present from the first dump onward, so no ID change is detected on subsequent dumps.
+	 */
+	private void clearClassDumpDirectoryIfReloadDetected(Dump dump) {
+		boolean reloadDetected = false;
+		for (ExecutionData data : dump.getStore().getContents()) {
+			Long previousId = previousClassIds.put(data.getName(), data.getId());
+			if (previousId != null && previousId != data.getId()) {
+				reloadDetected = true;
```
DreierF: This doesn't necessarily detect a reload. I could also just be two conflicting class files, i.e., a third-party dependency that is on the classpath in two different versions.

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -198,4 +208,40 @@ private void dumpReportUnsafe() {
...
+		if (!reloadDetected) {
+			return;
+		}
+
+		logger.info("Detected application server reload: class files were reloaded with different bytecode. "
+				+ "Clearing the class dump directory to remove stale class files from the previous deployment.");
+
+		Path classDumpDir = options.getClassDumpDirectory();
+		if (classDumpDir != null) {
+			FileSystemUtils.deleteRecursively(classDumpDir.toFile());
```
DreierF: Classes are only dumped to that directory by jacoco once. Some all class files from other jar files are also gone afterwards and hence will not produce any coverage anymore, but warnings, that the class files are missing.

DreierF (changes requested):
The approach looks good on the surface, but has some serious problems in some cases. I think we either need to more specifically detect the reload for individual jar files or at least ensure that if no `duplicates` is set, that the default `warn` does not crash the report generation, but just warn about the issue.

stahlbauer:
@DreierF Thanks for the review. I guess these are the pitfalls of Agentic Coding. Now a variant where primarily the duplication-detection is kept alive by sharing the visitor instance.

README.md
```diff
@@ -11,6 +11,86 @@
...
+│  │  Agent (Normal mode)         OR   TestwiseCoverageAgent         │     │
+│  │                                                                 │     │
+│  │  HTTP server (Jetty + Jersey) for control:                      │     │
+│  │    POST /dump ─── trigger coverage dump                         │     │
+│  │    POST /test/start, /test/end ─── per-test coverage            │     │
+│  └─────────────────────────────────────────────────────────────────┘     │
+│                                                                          │
+└──────────────────────────────────────────────────────────────────────────┘
+         │
+         │  Periodically (default: every 10 min) or on HTTP request
```
DreierF: I thought the default is `8h` 🤔

system-tests/reload-detection-test/src/test/kotlin/com/teamscale/systemtest/ReloadDetectionSystemTest.kt
```diff
@@ -0,0 +1,86 @@
...
+		// Simulate reload: load the same class with modified bytecode via a custom classloader.
+		// This causes JaCoCo's transformer to instrument the new version, creating an execution data
+		// entry with a different CRC64 for the same class name.
+		loadModifiedSystemUnderTest()
+
+		// Generate more coverage and dump again. Because the agent creates a fresh report generator
+		// (and thus a fresh CoverageBuilder) per dump, the changed class ID does not cause a crash.
+		SystemUnderTest().bar()
+		dumpCoverage(SystemTestUtils.AGENT_PORT)
+
```
DreierF: Can we also check that the log file does contain the expected warning? (we have some other tests that already check the log file)

system-tests/reload-detection-test/src/test/kotlin/com/teamscale/systemtest/ReloadDetectionSystemTest.kt
```diff
@@ -0,0 +1,86 @@
...
+ * old class ID from a previous dump. The fix is to create a fresh report generator per dump so the `CoverageBuilder`
+ * never carries state across dumps.
+ *
+ * The reload is simulated by loading a modified copy of the [SystemUnderTest] class through a custom classloader,
+ * which causes JaCoCo to create execution data with a different class ID for the same class name.
+ */
+class ReloadDetectionSystemTest {
+
+	@Test
+	fun dumpAfterClassReloadDoesNotCrash() {
```
DreierF: The test also passes on master without any changes to the agent, so it does not really reproduce the issue.

DreierF (changes requested):

agent/src/main/java/com/teamscale/jacoco/agent/options/JacocoAgentOptionsBuilder.java
```diff
@@ -35,11 +35,11 @@ public String createJacocoAgentOptions() throws AgentOptionParseException, IOExc
 		// Don't dump class files in testwise mode when coverage is written to an exec file
 		boolean needsClassFiles = agentOptions.mode == EMode.NORMAL || agentOptions.testwiseCoverageMode != ETestwiseCoverageMode.EXEC_FILE;
 		if (agentOptions.classDirectoriesOrZips.isEmpty() && needsClassFiles) {
-			Path tempDir = createTemporaryDumpDirectory();
-			tempDir.toFile().deleteOnExit();
-			builder.append(",classdumpdir=").append(tempDir.toAbsolutePath());
+			Path classDumpDirectory = createTemporaryDumpDirectory();
```
stahlbauer, line 38: Just a variable renaming

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
```diff
@@ -188,6 +184,13 @@ private void dumpReportUnsafe() {
 			return;
 		}
 
+		// Created fresh per dump so its CoverageBuilder does not carry class IDs from previous dumps.
+		// Without this, application server reloads (for example, JBoss) cause an IllegalStateException
+		// because the reloaded class has a different CRC64 than what CoverageBuilder stored earlier.
+		JaCoCoXmlReportGenerator generator = new JaCoCoXmlReportGenerator(
```
stahlbauer, line 187: Core of the bug fix

report-generator/src/main/kotlin/com/teamscale/report/jacoco/JaCoCoBasedReportGenerator.kt
```diff
@@ -17,7 +17,14 @@ import java.io.OutputStream
 /**
  * Base class for generating reports based on the binary JaCoCo exec dump files.
  *
- * It takes care of ignoring non-identical classes with the same fully qualified name and classes without coverage.
+ * JaCoCo's execution data only contains per-class boolean arrays indicating which probes fired. It does not contain
```
stahlbauer, line 20: Just better JavaDocs

README.md
```diff
@@ -11,6 +11,86 @@
...
+│  │  Agent (Normal mode)         OR   TestwiseCoverageAgent         │     │
+│  │                                                                 │     │
+│  │  HTTP server (Jetty + Jersey) for control:                      │     │
+│  │    POST /dump ─── trigger coverage dump                         │     │
+│  │    POST /test/start, /test/end ─── per-test coverage            │     │
+│  └─────────────────────────────────────────────────────────────────┘     │
+│                                                                          │
+└──────────────────────────────────────────────────────────────────────────┘
+         │
+         │  Periodically (default: every 480 min) or on HTTP request
```
stahlbauer, line 68: Now, the 8h are here. The cause was an inconsistency in CLAUDE.md

CLAUDE.md
```diff
@@ -51,7 +51,7 @@ Entry point: `PreMain.premain()` in `agent/src/main/java/com/teamscale/jacoco/ag
 4. Classes are instrumented at load time with coverage probes
 
 **Coverage modes:**
-- **Interval-based** (`Agent` class) - Periodically dumps coverage (default every 10 min)
+- **Interval-based** (`Agent` class) - Periodically dumps coverage (default every 480 min)
```
stahlbauer, line 54: This was inconsistent to the implementation

agent/src/main/java/com/teamscale/jacoco/agent/Agent.java
DreierF: IMHO this also just fixes the symptom instead of the core of the problem. I now looked a bit closer into the code: From my POV the problem is that the `EnhancedCoverageVisitor` delegates to `coverageVisitor`, but both have a different lifecycle. i.e., `EnhancedCoverageVisitor` is recreated for every dump, but the visitor keeps the previous data. Hence we don't properly warn (sole purpose of `EnhancedCoverageVisitor`) but instead crash in the JaCoCo code.

I would suggest to change `JaCoCoBasedReportGenerator` to not receive a `Visitor` instance, but rather a `Supplier<Visitor>` so  that we can create a new one in `convertSingleDumpToReport` pass that into `EnhancedCoverageVisitor`.

report-generator/src/test/kotlin/com/teamscale/report/jacoco/JaCoCoXmlReportGeneratorTest.kt
```diff
@@ -77,6 +77,91 @@ class JaCoCoXmlReportGeneratorTest : TestDataBase() {
...
+			)
+		}.isExactlyInstanceOf(IOException::class.java)
+			.hasCauseExactlyInstanceOf(IllegalStateException::class.java)
+	}
+
+	/**
+	 * Verifies that reusing a single [JaCoCoXmlReportGenerator] across multiple dumps crashes when the class files
+	 * change between dumps (for example, due to an application server reload). The internal [CoverageBuilder] retains
+	 * class IDs from the first dump and throws [IllegalStateException] when it encounters a different ID for the same
+	 * class name.
```
DreierF: We should not never fails hard in such a situation as you could also have conflicting class files at runtime for the same dump or collected in the class dump folder.

DreierF (changes requested):

report-generator/src/main/kotlin/com/teamscale/report/jacoco/JaCoCoBasedReportGenerator.kt
```diff
@@ -62,27 +78,48 @@ abstract class JaCoCoBasedReportGenerator<Visitor : ICoverageVisitor>(
...
 		codeDirectoriesOrArchives.forEach { file ->
-			FilteringAnalyzer(store, EnhancedCoverageVisitor(), locationIncludeFilter, logger)
+			FilteringAnalyzer(store, visitor, locationIncludeFilter, logger)
 				.analyzeAll(file)
 		}
 	}
 
-	private inner class EnhancedCoverageVisitor : ICoverageVisitor {
+	/**
+	 * Wrapper around the actual coverage visitor (typically JaCoCo's [org.jacoco.core.analysis.CoverageBuilder] or
```
stahlbauer, line 115: Some more docs that describe the purpose of this.

report-generator/src/main/kotlin/com/teamscale/report/jacoco/JaCoCoBasedReportGenerator.kt
```diff
@@ -62,27 +78,48 @@ abstract class JaCoCoBasedReportGenerator<Visitor : ICoverageVisitor>(
...
+	 * application server reload) would bypass our duplicate handling and hit `CoverageBuilder` directly, which always
+	 * throws `IllegalStateException` regardless of the configured [duplicateClassFileBehavior].
 	 */
 	@Throws(IOException::class)
-	private fun analyzeStructureAndAnnotateCoverage(store: ExecutionDataStore) {
+	private fun analyzeStructureAndAnnotateCoverage(store: ExecutionDataStore, coverageVisitor: Visitor) {
+		val visitor = EnhancedCoverageVisitor(coverageVisitor)
 		codeDirectoriesOrArchives.forEach { file ->
-			FilteringAnalyzer(store, EnhancedCoverageVisitor(), locationIncludeFilter, logger)
+			FilteringAnalyzer(store, visitor, locationIncludeFilter, logger)
```
stahlbauer, line 109: The same class can appear in different paths/jars/wars.

report-generator/src/main/kotlin/com/teamscale/report/jacoco/JaCoCoBasedReportGenerator.kt
```diff
@@ -43,9 +58,10 @@ abstract class JaCoCoBasedReportGenerator<Visitor : ICoverageVisitor>(
 	fun convertSingleDumpToReport(dump: Dump, outputFilePath: File): CoverageFile {
 		val coverageFile = CoverageFile(outputFilePath)
 		val mergedStore = dump.store
-		analyzeStructureAndAnnotateCoverage(mergedStore)
+		val coverageVisitor = coverageVisitorSupplier()
```
stahlbauer, line 61: For each dump call, a new visitor is created so that classes are not left out accidentally in dump calls that follow.

PR #882, opened 2026-04-14
TS-43260 Profiler crash fix in case of invalid config options

stahlbauer:
Addresses issue [TS-43260](https://cqse.atlassian.net/browse/TS-43260)

- [ ] CHANGELOG.md updated


agent/src/main/java/com/teamscale/jacoco/agent/PreMain.java
```diff
@@ -97,7 +97,9 @@ public static void premain(String options, Instrumentation instrumentation) thro
 				agentOptions.configurationViaTeamscale.unregisterProfiler();
 			}
 
-			throw e;
+			// Don't crash the profiled application due to a configuration error
+			// (see TS-43260). The error has already been logged in getAndApplyAgentOptions.
+			return;
```
DreierF, line 102: The TI team discussed the topic and came to the (surprising) decision, that the agent should **never** crash the profiled application even if the agent is configured incorrectly. Can you please adjust the other places where we currently re-throw exceptions, document this decision in the code and adjust the tests if necessary.

stahlbauer: Adjusted.

PR #883, opened 2026-04-16
TS-46051 Switch from docker to oci for publishing

DreierF:
Addresses issue [TS-46051](https://cqse.atlassian.net/browse/TS-46051)

- [x] Teamscale documentation updated in case of user-visible changes
- [x] CHANGELOG.md updated


agent/src/docker/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+if [ -e /transfer ]; then
```
stahlbauer, line 5: A comment would help to understand what is here happening why.

agent/build.gradle.kts
```diff
@@ -101,3 +104,57 @@ distributions {
...
+			}
+			layer("entrypoint") {
+				contents {
+					from("src/docker/entrypoint.sh") {
+						filePermissions = "755".toInt(8)
+					}
+				}
+			}
+		}
+		specificPlatform(platform("linux", "amd64"))
```
stahlbauer, line 148: So I guess we do not support old 32bit architectures at all anymore. (Which is fine. Just checking.)

DreierF: No, also didn't do before, but 32bit containers are also a very rare thing.

PR #912, opened 2026-05-28
TS-38628 Codebase Cleanup

Avanatiker:
DEFAULT_EXCLUDES — added com.teamscale.*:kotlin.* to prevent agent self-instrumentation
locationIncludeFilter — computed getter → by lazy (was recompiling WildcardMatcher on every access)
ClasspathWildcardIncludeFilter.getClassName() — cached Regex, JavaNames, and results via ConcurrentHashMap
AgentResource / TestwiseCoverageResource / ResourceBase — mutable companion object state → constructor injection
RevisionInfo / ERevisionType — enum + nullable field → sealed class with Commit/Revision variants
ITestDescriptorResolver — Optional<String> return types → String? throughout 10 files
TeamscaleCredentials, FileCoverage, PathCoverage — class → data class
Gradle plugin — suppressed deprecation warnings, added missing jacoco.core dependency
Kotlin version — 2.3.21 → 2.3.20 (matches Gradle's embedded version)


agent/src/main/kotlin/com/teamscale/jacoco/agent/logging/LogToTeamscaleAppender.kt
```diff
@@ -27,41 +27,31 @@ class LogToTeamscaleAppender : AppenderBase<ILoggingEvent>() {
...
-	 * Buffer for unsent logs. We use a set here to allow for removing entries fast after sending them to Teamscale was
-	 * successful.
+	 * Thread-safe buffer for unsent logs. Using [ConcurrentLinkedQueue] for lock-free producer-consumer access.
 	 */
-	private val logBuffer = LinkedHashSet<ProfilerLogEntry>()
+	private val logBuffer = ConcurrentLinkedQueue<ProfilerLogEntry>()
 
 	/** Scheduler for sending logs after the configured time interval  */
 	private val scheduler: ScheduledExecutorService = Executors.newScheduledThreadPool(1) { r ->
-		// Make the thread a daemon so that it does not prevent the JVM from terminating.
```
DreierF, line 37: I'd like to keep this comment as it might not be entirely obvious for everyone.

agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitPropertiesLocatingTransformer.kt
```diff
@@ -26,12 +28,17 @@ class GitPropertiesLocatingTransformer(
 		classFileContent: ByteArray?
 	): ByteArray? {
 		if (protectionDomain == null) {
-			// happens for e.g. java.lang. We can ignore these classes
```
DreierF, line 29: I think this comment still adds useful information.

agent/src/main/kotlin/com/teamscale/jacoco/agent/logging/LogToTeamscaleAppender.kt
```diff
@@ -27,41 +27,31 @@ class LogToTeamscaleAppender : AppenderBase<ILoggingEvent>() {
...
+		)
 	}
 
 	override fun append(eventObject: ILoggingEvent) {
-		synchronized(logBuffer) {
-			logBuffer.add(formatLog(eventObject))
-			if (logBuffer.size >= BATCH_SIZE) flush()
-		}
+		logBuffer.add(formatLog(eventObject))
+		if (logBuffer.size >= BATCH_SIZE) flush()
```
DreierF: `size` of  `ConcurrentLinkedQueue` is `O(n)`. Considering that logs might have a lot of entries this might be a performance problem.

report-generator/src/main/kotlin/com/teamscale/jacoco.agent.options/FilePatternResolver.kt
```diff
@@ -0,0 +1,183 @@
...
+			}
+
+		/**
+		 * Resolves the pattern. The results can be retrieved via [singlePath] or [allMatchingPaths].
+		 */
+		fun resolve(): FilePatternResolverRun {
+			val pathRegex = AntPatternUtils.convertPattern(suffixPattern, false)
+
+			try {
+				matchingPaths = basePath.walk().filter {
```
DreierF: I'm not sure whether we need that or not but the previous code also listed directories. This one only lists files. Maybe we should add `PathWalkOption.INCLUDE_DIRECTORIES` to be on the safe side.

agent/src/main/kotlin/com/teamscale/jacoco/agent/logging/LogToTeamscaleAppender.kt
```diff
@@ -73,50 +63,35 @@ class LogToTeamscaleAppender : AppenderBase<ILoggingEvent>() {
...
+					logBuffer.addAll(batch)
+				} else {
+					addStatus(ErrorStatus("Sending logs to Teamscale failed: ${e.message}", this, e))
+					logBuffer.addAll(batch)
 				}
-			}))
+			}
 		}
+		future.whenComplete { _, _ -> activeLogFlushes.remove(future) }
+		activeLogFlushes.add(future)
```
DreierF: This is a silent race condition. If the `future` is already completed, the `whenComplete` callback will be executed synchronously. So you will end up with a stale future in `activeLogFlushes`. 
```suggestion
		activeLogFlushes.add(future)
		future.whenComplete { _, _ -> activeLogFlushes.remove(future) }
```

agent/src/main/kotlin/com/teamscale/jacoco/agent/options/AgentOptions.kt
```diff
@@ -695,7 +696,7 @@ open class AgentOptions(private val logger: ILogger) {
 		 * debugging traces easier and reduces trace size and warnings about unmatched classes in Teamscale.
 		 */
 		const val DEFAULT_EXCLUDES =
-			"shadow.*:com.sun.*:sun.*:org.eclipse.*:org.junit.*:junit.*:org.apache.*:org.slf4j.*:javax.*:org.gradle.*:java.*:org.jboss.*:org.wildfly.*:org.springframework.*:com.fasterxml.*:jakarta.*:org.aspectj.*:org.h2.*:org.hibernate.*:org.assertj.*:org.mockito.*:org.thymeleaf.*"
+			"com.teamscale.*:kotlin.*:shadow.*:com.sun.*:sun.*:org.eclipse.*:org.junit.*:junit.*:org.apache.*:org.slf4j.*:javax.*:org.gradle.*:java.*:org.jboss.*:org.wildfly.*:org.springframework.*:com.fasterxml.*:jakarta.*:org.aspectj.*:org.h2.*:org.hibernate.*:org.assertj.*:org.mockito.*:org.thymeleaf.*"
```
DreierF: This will no longer allow us to use the Java profiler for the Teamscale code itself.

agent/src/main/kotlin/com/teamscale/jacoco/agent/logging/LogToTeamscaleAppender.kt
```diff
@@ -73,50 +63,35 @@ class LogToTeamscaleAppender : AppenderBase<ILoggingEvent>() {
...
+
+		val future = CompletableFuture.runAsync {
+			val client = teamscaleClient ?: return@runAsync
+
+			try {
+				val response = client.postProfilerLog(profilerId!!, batch).execute()
+				check(response.isSuccessful) { "Failed to send log: HTTP error code : ${response.code()}" }
+			} catch (e: Exception) {
+				if (e is ConnectException) {
+					logBuffer.addAll(batch)
```
DreierF: This will change the order of the log entries as they arrive in Teamscale. We can probably work around it with `addFirst`? (Same below)

Also should be move this line out of the if statement and invert the if condition?

report-generator/src/main/kotlin/com/teamscale/jacoco/agent/options/ClasspathUtils.kt
DreierF, line 1: The file path `teamscale/jacoco.agent.options` should be `teamscale/jacoco/agent/options` 😉

gradle/libs.versions.toml
```diff
@@ -89,6 +89,6 @@ gitProperties = { id = "com.gorylenko.gradle-git-properties", version = "3.0.3"
 mavenPluginDevelopment = { id = "org.gradlex.maven-plugin-development", version = "1.0.3" }
 shadow = { id = "com.gradleup.shadow", version = "9.4.2" }
 oci = { id = "io.github.sgtsilvio.gradle.oci", version = "0.27.0" }
-kotlinJvm = { id = "org.jetbrains.kotlin.jvm", version = "2.3.21" }
+kotlinJvm = { id = "org.jetbrains.kotlin.jvm", version = "2.3.20" }
```
DreierF: Would be good to also update the renovate configuration because otherwise the update would directly be reapplied after merging this.

agent/src/main/kotlin/com/teamscale/jacoco/agent/LenientCoverageTransformer.kt
```diff
@@ -2,25 +2,23 @@ package com.teamscale.jacoco.agent
...
- * A class file transformer which delegates to the JaCoCo [org.jacoco.agent.rt.internal_29a6edd.CoverageTransformer] to do the actual instrumentation,
+ * A class file transformer which delegates to the JaCoCo [CoverageTransformer] to do the actual instrumentation,
  * but treats instrumentation errors e.g. due to unsupported class file versions more lenient by only logging them, but
  * not bailing out completely. Those unsupported classes will not be instrumented and will therefore not be contained in
  * the collected coverage report.
  */
 class LenientCoverageTransformer(
 	runtime: IRuntime?,
-	options: AgentOptions,
+	options: org.jacoco.agent.rt.internal_29a6edd.core.runtime.AgentOptions,
```
DreierF: Why not keep the import?

report-generator/src/main/kotlin/com/teamscale/report/testwise/model/RevisionInfo.kt
```diff
@@ -1,35 +1,35 @@
...
 
 /** Revision information necessary for uploading reports to Teamscale.  */
-class RevisionInfo : Serializable {
-	/** The type of revision information.  */
-	val type: ERevisionType
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
+@JsonSubTypes(
+	JsonSubTypes.Type(value = RevisionInfo.Commit::class, name = "COMMIT"),
+	JsonSubTypes.Type(value = RevisionInfo.Revision::class, name = "REVISION")
+)
```
DreierF, line 14: We might want to add a short changelog entry about this as it changes the REST API of the agent.

DreierF (changes requested):

report-generator/src/main/kotlin/com/teamscale/jacoco/agent/options/ClasspathUtils.kt
Avanatiker: Damn intellij wtf xD

agent/src/main/kotlin/com/teamscale/jacoco/agent/LenientCoverageTransformer.kt
Avanatiker: I had another AgentOptions (ours) in the class before so it was colliding but i changed it back

agent/src/main/kotlin/com/teamscale/jacoco/agent/logging/LogToTeamscaleAppender.kt
```diff
@@ -13,11 +13,14 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
+import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.delay
```
DreierF: This and a few of the following imports are unused.

agent/src/main/kotlin/com/teamscale/jacoco/agent/logging/LogToTeamscaleAppender.kt
```diff
@@ -8,112 +8,93 @@ import ch.qos.logback.core.status.ErrorStatus
...
+	/**
+	 * Collector coroutine: receives entries from [logChannel], batches them up to [BATCH_SIZE], and sends them
+	 * via [sendBatch]. If no entries arrive within [FLUSH_INTERVAL], a new cycle begins (no empty batch is sent).
+	 * Exits when the scope is cancelled or the channel is closed and empty.
+	 */
+	private suspend fun collectAndSend() {
+		while (currentCoroutineContext().isActive) {
+			val batch = buildList {
+				val entry = withTimeoutOrNull(FLUSH_INTERVAL) {
+					logChannel.receive()
```
DreierF: Should we use `receiveCatching` here? According to Claude this would otherwise throw an exception on shutdown even when it terminated "normally".

DreierF (changes requested):

PR #962, opened 2026-08-20
TS-47508 Accept a folder for the git-properties-jar option

karottenreibe:
## What

`git-properties-jar` now also accepts a folder, which is searched for `git.properties` files. Previously a folder was rejected with a warning and the option was dropped, so deployments that keep `git.properties` in a resources folder (e.g. `/app/resources` in a container image) instead of inside the profiled archive got no commit auto-detection at all.

## How

The search machinery already handled both shapes: `GitPropertiesLocatorUtils.findGitPropertiesInFile` branches into an archive or a folder search, and the folder branch is what the automatic detection already uses for exploded class folders (`BOOT-INF/classes`, JBoss vfs temp folders). The obstacles were only:

- the `!isFile()` rejection in `AgentOptionsParser.getGitPropertiesJarFile` (now `getGitPropertiesFileOrFolder`), and
- the three call sites in `AgentOptions` hardcoding `isJarFile = true`, which now derive it from `isDirectory()`. `ArtifactoryConfig.parseGitProperties` got the corresponding parameter.

A non-existent path still logs a warning and lets the application continue.

## Decisions

- **Folder recursion semantics unchanged:** the folder tree is always walked for `git.properties`, while `search-git-properties-recursively` only controls whether archives *inside* the folder are opened too. Making the flag also gate subfolders would change the existing automatic detection for exploded class folders, where `git.properties` usually sits in a subfolder — a breaking change with no upside here.
- **Option name kept:** no new alias and no deprecation, so an option named `...-jar` now also takes a folder. Documented in the KDoc, the log messages and the CHANGELOG.

## Drive-by fix

The inconsistency warning in `GitSingleProjectPropertiesLocator` pointed users at `teamscale-git-properties-jar`, which was replaced by `git-properties-jar` in 31.0.0. It now references the current option constant. The Artifactory "found no git.properties" message likewise named the deprecated `artifactory-git-properties-jar`.

## Tests

- `AgentOptionsTest.testGitPropertiesJarDoesNotAcceptFolders` replaced by `testGitPropertiesJarAcceptsFolders`.
- `DelayedCommitDescriptorRetrievalTest` now covers both an archive and a folder search root.
- Folder scanning itself was already covered by `GitPropertiesLocatorTest.testReadingGitPropertiesInJarFileNestedInFolder`.
- `./gradlew :agent:test` passes.

Manual verification against a real Teamscale/Artifactory upload is still pending.

## Out of scope

Testwise mode error messages list `git-properties-jar` as a way to supply a commit, but no testwise code path reads `gitPropertiesJar`. That pre-existing gap is untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitSingleProjectPropertiesLocator.kt
+++ b/agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitSingleProjectPropertiesLocator.kt
@@ -1,6 +1,7 @@
 package com.teamscale.jacoco.agent.commit_resolution.git_properties
 
 import com.teamscale.jacoco.agent.logging.LoggingUtils.getLogger
+import com.teamscale.jacoco.agent.options.AgentOptions
 import com.teamscale.jacoco.agent.upload.delay.DelayedUploader
 import com.teamscale.jacoco.agent.util.DaemonThreadFactory
 import java.io.File
@@ -10,8 +11,8 @@ import java.util.concurrent.Executor
 import java.util.concurrent.Executors
 
 /**
- * Searches a Jar/War/Ear/... file for a git.properties file in order to enable upload for the commit described therein,
- * e.g. to Teamscale, via a [DelayedUploader].
+ * Searches a Jar/War/Ear/... file or a folder for a git.properties file in order to enable upload for the commit
+ * described therein, e.g. to Teamscale, via a [DelayedUploader].
  */
 class GitSingleProjectPropertiesLocator<T>(
 	private val uploader: DelayedUploader<T>,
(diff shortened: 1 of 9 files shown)


agent/src/main/kotlin/com/teamscale/jacoco/agent/commit_resolution/git_properties/GitSingleProjectPropertiesLocator.kt
```diff
@@ -1,6 +1,7 @@
 package com.teamscale.jacoco.agent.commit_resolution.git_properties
 
 import com.teamscale.jacoco.agent.logging.LoggingUtils.getLogger
+import com.teamscale.jacoco.agent.options.AgentOptions
```
karottenreibe, line 4: I think the architecture needs to be adjusted to make the option component public.
Already there's tons of dependencies on this and anyone wanting to include an option name in an error message needs to reference constants from those classes. 

I don't like the alternative of hard-coding the option name here. And I think having a separate arch component just for option names is also a bit overkill. WDYT @DreierF ?

DreierF: 👍 Makes sense to me

DreierF: I adjusted the architecture on master

PR #965, opened 2026-08-31
TS-46357 Collect the tests of a `@ParameterizedClass`

DreierF:
Addresses issue [TS-46357](https://cqse.atlassian.net/browse/TS-46357)

- [ ] Teamscale documentation updated in case of user-visible changes
- [x] CHANGELOG.md updated

### DIFF
--- a/agent/src/main/kotlin/com/teamscale/jacoco/agent/testimpact/CoverageViaHttpStrategy.kt
+++ b/agent/src/main/kotlin/com/teamscale/jacoco/agent/testimpact/CoverageViaHttpStrategy.kt
@@ -28,8 +28,8 @@ class CoverageViaHttpStrategy(
 		val builder = TestInfoBuilder(test)
 		val dump = controller.dumpAndReset()
 		reportGenerator.updateClassDirCache()
-		reportGenerator.convert(dump)?.let { builder.setCoverage(it) }
-		builder.setExecution(testExecution)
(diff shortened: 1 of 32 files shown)


agent/src/main/kotlin/com/teamscale/jacoco/agent/testimpact/CoverageViaHttpStrategy.kt
```diff
@@ -28,8 +28,8 @@ class CoverageViaHttpStrategy(
 		val builder = TestInfoBuilder(test)
 		val dump = controller.dumpAndReset()
 		reportGenerator.updateClassDirCache()
-		reportGenerator.convert(dump)?.let { builder.setCoverage(it) }
-		builder.setExecution(testExecution)
+		reportGenerator.convert(dump)?.let { builder.addCoverage(it) }
+		builder.addExecution(testExecution)
```
Stef2k16, line 32: This method could be called multiple times via the `/test/end` endpoint (e.g, if the caller has to retry for whatever reason). We might not want to add, but replace in such cases?

DreierF: The builder is created per request and the result directly returned via HTTP again, so there is nothing to replace.

PR #976, opened 2026-09-17
Fix the Windows system tests under Java 25

DreierF:
Supersedes #975: contains the Java 25 bump plus the two changes that make it pass on Windows.

### What broke

`system-tests/gradle-multi-module/gradle-project/gradle/gradle-daemon-jvm.properties` pinned the fixture's **Gradle daemon** to Java 21. That is stricter than the Java 21 toolchain the fixture's projects compile against, because the daemon's JVM is chosen by the launcher before the settings script is evaluated — so the foojay toolchain repository declared there does not exist yet and the JDK cannot be provisioned either.

While the runners ran Gradle on Java 21 the launcher's own JVM satisfied the requirement. On Java 25 it no longer does, and only the Linux runners keep their JDKs where the launcher detects them (`/usr/lib/jvm`); on Windows they live in the tool cache, so every nested build failed with

```
Unable to download toolchain matching the requirements ({languageVersion=21, ...}) from 'null', due to:
No defined toolchain download url for WINDOWS on x86_64 architecture.
```

The file is the only one of its kind in the repository and arrived as a stray in an unrelated commit (5a6b8cf4); no test asserts anything about the daemon's JVM. The projects keep their Java 21 toolchain, which resolves normally once the build is running.

### Why it took a CI round to find

A nested Gradle build that fails while configuring writes nothing but the single-use daemon notice to stdout and reports the failure on stderr. `runGradle` handed that result back to the caller, so the tests failed on `Expected exactly one session, but got 0` and the cause never reached the test report. It now reports a failed build instead, which is what made the error above visible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DreierF:
Superseded: the two fixes now live on renovate/java-jdk-25.x (#975), which already carries the Java 25 bump.

